PR #8, opened 2026-04-10
Add PlanetScale engine implementation

aparajon:
## Summary

Replaces the stub PlanetScale engine (all methods returning "not implemented") with the full implementation backed by the PlanetScale SDK and Vitess online DDL.

The engine implements the complete deploy lifecycle: Plan diffs schema files against live databases using Spirit's PlanChanges (shared diff+lint pipeline with the Spirit engine), Apply creates branches and deploy requests with DDL applied via MySQL connections, and Progress tracks per-shard Vitess migration status via SHOW VITESS_MIGRATIONS. VSchema changes are diffed using protobuf normalization and applied alongside DDL.

Key design decisions:
- Concurrent keyspace DDL application via errgroup (capped at 6 parallel keyspaces) with retries (3 attempts, exponential backoff + jitter)
- Crash recovery via `ApplyRequest.OnStateChange` callback — the engine persists ResumeState at key milestones (branch creation, deploy request creation) so the tern layer can resume after worker crashes without starting over
- On resume with a partially-applied branch, diffs the branch against desired schema to apply only the missing DDL rather than re-applying everything
- Safe migrations check before branch creation prevents confusing failures
- API call timeouts (10-30s) and branch password TTL (1h) for resilience
- Engine-specific credentials (TLS config) use the generic Credentials.Metadata map per the engine-agnostic interface guideline
- Uses slog directly instead of per-instance logger
- `Cancelled` state constant added to pkg/state to match PlanetScale API's deployment states
- Default throttle volume (2) maps to 0.85 ratio, clamped to [0.0, 0.95]

### DIFF
--- a/pkg/engine/planetscale/planetscale.go
+++ b/pkg/engine/planetscale/planetscale.go
@@ -1467,6 +1467,7 @@ const DefaultVolume int32 = 2
 
 // volumeToThrottleRatio converts volume (1-11) to a PlanetScale throttle ratio.
 // Lower volume = more throttling. DefaultVolume (2) maps to 0.85.
+// See engine.VolumeRequest for how volume semantics differ between engines.
 var volumeThrottleMap = [12]float64{
 	0:  0.95, // unused (volume is 1-indexed)
 	1:  0.95, // max throttle
--- a/pkg/engine/planetscale/planetscale.go
+++ b/pkg/engine/planetscale/planetscale.go
@@ -84,7 +84,7 @@
 //
 //  2. Vitess migrations via SHOW VITESS_MIGRATIONS: per-table, per-shard row counts,
 //     ETA, progress %, migration context, cutover attempts, throttle reasons.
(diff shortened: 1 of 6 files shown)


pkg/engine/planetscale/planetscale.go
```diff
@@ -384,76 +386,1778 @@ package planetscale
...
+	3:  0.75,
+	4:  0.65,
+	5:  0.55,
+	6:  0.45,
+	7:  0.35,
+	8:  0.25,
+	9:  0.15,
+	10: 0.05,
+	11: 0.0, // no throttle
+}
```
morgo, line 1484: This might not map perfectly to spirit because no throttle = 1 thread, versus spirit volume 3+ is usually 2+ threads. Having said that the instances are usually smaller so by percentage of resources 1 thread is more.

Not a blocker, just wonder if there's a quick note we can add somewhere to explain that these values are not really transferable between engines, they are really apples to oranges.

morgo (dismissed):

morgo (approved):

PR #13, opened 2026-04-15
Support structured storage connection fields

aparajon:
StorageConfig now accepts individual connection fields (host, port, username, password, database) as an alternative to a single DSN string. Each field supports secret resolution prefixes (env:, file:, secretsmanager:), and SchemaBot assembles the DSN at startup using the MySQL driver's FormatDSN for safe escaping.

This is useful for Kubernetes deployments where credentials are mounted as individual files via External Secrets Operator — for example, when the database password is managed by RDS in a separate secret. ESO syncs both secrets into a single k8s secret volume, and SchemaBot reads all fields uniformly as files.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -36,14 +36,57 @@ tern_deployments:
     production: "tern1-production:9090"
 ```
 
+## Storage Connection
+
+The `storage` section configures SchemaBot's internal database. You can provide the connection as a single DSN string or as individual fields.
+
+### DSN (default)
+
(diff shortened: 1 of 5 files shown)


morgo (dismissed):

aparajon:
Closing in favor of an ESO template approach — ESO constructs the DSN from the connection secret and RDS-managed password, so SchemaBot just reads a single `file:/secrets/db/dsn`. No app code changes needed.

aparajon:
Open to reviving this if we find that we need this feature

PR #34, opened 2026-04-23
Prefer instant, fix cutover progress, skip-revert, TUI fixes, and symlink support

aparajon:
Fixes issues discovered during real Vitess applies and improves CLI rendering consistency.

- **Prefer instant DDL** — instant DDL is now used whenever PlanetScale reports eligibility, regardless of `skip_revert`. The TUI shows "⚡ Applying instantly..." during the operation.
- **LocalScale instant detection** — tests ALGORITHM=INSTANT on a scratch database during deploy request preparation, matching real PlanetScale behavior. Works regardless of how DDL was applied to the branch.
- **Clamp progress to 100% at cutover** — shard and table progress no longer show 99% when the copy phase is done.
- **Skip-revert timeout** — the TUI "Finalizing" spinner times out after 10s and allows retry via Enter.
- **ESC detach during branch creation** — the TUI now shows an ESC hint during setup phases.
- **Follow symlinks in schema directories** — keyspace subdirectories that are symlinks were silently skipped during plan.
- **Unified plan/progress rendering** — plan output now uses the same keyspace headers (── name ──) and DDL formatting as the progress view.
- **Better error messages** — CLI errors shown in red; HTML auth pages detected with visible text instead of cryptic JSON parse errors.
- **Store IsInstant in VitessApplyData** — the instant DDL flag persists across server restarts and progress queries.


pkg/cmd/client/request.go
```diff
@@ -123,11 +123,66 @@ func doPostInto(endpoint, path string, body any, result any) error {
...
+		if text != "" {
+			return fmt.Errorf("unexpected response from server (received HTML, expected JSON):\n  %s", text)
+		}
+		return fmt.Errorf("unexpected response from server (received HTML, expected JSON)")
+	}
+	return fmt.Errorf("parse response: %w", parseErr)
+}
+
+// stripHTMLTags removes HTML tags and style/script content, returning only visible text.
+func stripHTMLTags(s string) string {
```
morgo: Is there a library you could use for this instead. It will likely be more robust.

pkg/localscale/helpers.go
```diff
@@ -569,6 +569,131 @@ func normalizeJSON(data []byte) string {
...
+	}
+	rest := strings.TrimSpace(trimmed[len("ALTER TABLE "):])
+	if strings.HasPrefix(rest, "`") {
+		if end := strings.Index(rest[1:], "`"); end >= 0 {
+			tableName := rest[1 : end+1]
+			afterTable := strings.TrimSpace(rest[end+2:])
+			return tableName, afterTable
+		}
+		return "", ""
+	}
```
morgo: If these were pre-parsed it should be OK to check exactly like this, but at the same time it is doing strings.TrimSpace.

We should really be using the spirit statement parsing here - if there is functionality you are missing, we can add it.

morgo (approved):

PR #43, opened 2026-04-27
fix(github): fix rollback-confirm losing PR comments and check run updates

aparajon:
When a rollback completes via PR comment, the progress and summary comments should appear on the PR and the status check should update. This was silently broken for local mode — the apply executed but no feedback was posted.

The root cause: the background progress watcher read GitHub context (repo, PR number, installation ID) from the apply record in storage. In local mode, these fields were never set because the proto interface doesn't carry GitHub-specific fields. The watcher ended up posting to invalid API URLs.

The fix passes these values directly from the webhook handler where they're already available, rather than relying on the storage record. Adds an end-to-end test for the full rollback-confirm flow that will catch this regression if it recurs.

### DIFF
--- a/pkg/webhook/webhook_e2e_test.go
+++ b/pkg/webhook/webhook_e2e_test.go
@@ -75,7 +75,7 @@ import (
 	"testing"
 	"time"
 
-	_ "github.com/go-sql-driver/mysql"
+	mysql "github.com/go-sql-driver/mysql"
 	gh "github.com/google/go-github/v68/github"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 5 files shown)


pkg/webhook/webhook_e2e_test.go
```diff
@@ -1659,6 +1659,110 @@ func TestE2ERollbackConfirmNoLock(t *testing.T) {
...
+// on the correct PR. This catches regressions where watchApplyProgress loses the
+// repo/PR/installationID context and fails to post comments.
+func TestE2ERollbackConfirmExecutesAndPostsComments(t *testing.T) {
+	dbName := "webhook_rbconfirm_exec"
+	svc := setupE2EService(t, dbName)
+	ctx := t.Context()
+
+	// Step 1: Create initial table
+	appDSN := strings.Replace(e2eTargetDSN, "/target_test", "/"+dbName, 1) + "&multiStatements=true"
+	db, err := sql.Open("mysql", appDSN)
```
morgo: I would prefer we convert to a mysql.Config, then change the dbName, but it's a test, so I guess its OK.

morgo (dismissed):
Small nit, optional to fix, otherwise LGTM.

morgo (approved):

PR #55, opened 2026-04-29
refactor: use Spirit utilities and typed constants consistently

aparajon:
Replace manual DDL parsing, raw state string comparisons, and ad-hoc patterns with their canonical Spirit/SchemaBot equivalents.

- **DDL type threading**: `engine.TableChange.Operation` is now `statement.StatementType` instead of a string. Spirit's typed constants (`StatementCreateTable`, etc.) flow through the engine layer — string conversion only happens at storage/proto boundaries via `ddl.StatementTypeToOp`/`OpToStatementType`.
- **DDL classification**: `addAlgorithmInstant` uses `statement.New()` instead of hand-parsing. `namespaceForTable` uses `ddl.ClassifyStatementAST()` instead of substring matching. `countStatementTypes` and `executeMigration` use `statement.Classify()` with typed constants.
- **State comparisons**: Raw `==`, `strings.EqualFold`, and `strings.ToUpper` state checks replaced with `state.IsState()` across CLI, TUI, progress templates, and tests.
- **Concurrency**: Manual `wg.Add/Done` replaced with `wg.Go()` in `FetchSchemaFilesOptimized`.
- **CLAUDE.md**: Added guidance to prefer Spirit's typed constants for DDL classification and push them to the edge.


aparajon:
**Suggestion for this PR**: Add constants for webhook command action strings (e.g., `"apply"`, `"plan"`, `"apply-confirm"`, `"unlock"`, etc.). These are currently raw strings in `commands.go`, `issue_comment.go`, and tests — same pattern we fixed for check run statuses/conclusions in #38. ~15-20 occurrences, mechanical find-and-replace.

morgo (approved):

PR #59, opened 2026-04-30
fix(vitess): skip VSchema validation in post-apply check

aparajon:
## Why
Follow-up to #58. VSchema applied via `UpdateKeyspaceVSchema` is read back via `GetKeyspaceVSchema`, which returns stale data due to the same PlanetScale API caching that affected DDL validation. VSchema cannot be fetched via MySQL.

## What
- Skip VSchema validation in the post-apply DDL check (DDL validation via MySQL remains)

## Risk Assessment
Low — VSchema is still validated during Plan. The post-apply check now only validates DDL, which is the higher-risk surface.

Generated with [Claude Code](https://claude.ai/code)


aparajon:
Closing for #60

PR #76, opened 2026-05-08
fix(github): post aggregate check runs for all PR types

aparajon:
## Summary

Ensure SchemaBot always posts aggregate check runs so branch protection required checks are satisfied on every PR type:

- **No schema files**: post passing aggregates immediately
- **Schema files without schemabot.yaml** (not onboarded): post passing aggregates with "No managed schema changes"
- **Schema files, wrong environment**: post passing aggregate for environments this instance doesn't manage (e.g., production instance sees a staging-only database)
- **Plan errors** (e.g., database not configured): post failing aggregates so branch protection shows a red check instead of waiting forever
- **Single-instance mode**: post global "SchemaBot" aggregate (not just per-environment)
- **Race prevention**: skip passing aggregates when existing check records are found (cleanupStaleChecks handles that case)
- **Short-circuit**: skip API calls when existing aggregate already passes for current SHA

### DIFF
--- a/pkg/webhook/check_runs.go
+++ b/pkg/webhook/check_runs.go
@@ -463,6 +463,102 @@ func (h *Handler) upsertAggregateCheckRun(
 		"per_database_checks", len(dbChecks))
 }
 
+// postPassingAggregates posts a passing aggregate check for each allowed environment.
+// Called when this instance has no work to do for a PR — either because the PR doesn't
+// touch schema files, or because the databases don't have environments this instance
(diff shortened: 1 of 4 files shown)


Kiran01bm (approved):

Kiran01bm:
lgtm - but i have a related improvement here if you'd like to consider - https://github.com/block/schemabot/pull/77

morgo (approved):

PR #113, opened 2026-05-18
feat: add `ListMysqlDatabases` endpoint 

VoLKyyyOG:
Create new endpoint for use by metanexus.
This lists all mysql databases.

ref: https://linear.app/squareup/issue/DFNS-445/create-schemabot-endpoint-listmysqldatabases

### DIFF
--- a/pkg/api/README.md
+++ b/pkg/api/README.md
@@ -54,6 +54,7 @@ Lazy creation means a connection failure to one database doesn't block startup o
 | GET | `/api/status` | `handleStatus` | All active schema changes |
 | GET | `/api/history/{database}` | `handleDatabaseHistory` | Apply history for a database |
 | GET | `/api/databases/{database}/environments` | `handleDatabaseEnvironments` | List environments |
+| GET | `/api/mysql/databases` | `handleListMysqlDatabases` | List configured MySQL databases |
 | GET | `/api/logs/{database}` | `handleLogs` | Apply logs for a database |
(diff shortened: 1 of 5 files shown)


aparajon:
Triage review of this draft, which has been open a while — sharing an honest assessment of where it stands relative to the current codebase:

1. **The API shape conflicts with the engine-agnostic surface rule** — `pkg/apitypes` and the route are MySQL-specific (`/api/mysql/databases`, `MysqlDatabaseResponse`), but the API layer is required to stay engine-agnostic; a database listing should be a generic `GET /api/databases` registry view that works for all engine types, with engine-specific data in metadata.

2. **Correctness: the `Deployment` field is populated with the database name** — deployment is per-environment configuration, so consumers would receive fabricated routing metadata.

3. **Staleness** — the branch is ~174 commits behind main and predates the per-environment `deployments` map, which the response type can't represent. There's also an exported handler wrapper that's a pure delegation (the repo avoids those).

4. **PR description** — the body references an internal service and an internal issue tracker link; per repo policy those need to be removed from public PR descriptions regardless of outcome.

Given the schema-pull/onboarding primitives now landing (#313), the recommendation is to close this and re-approach the need as an engine-agnostic database listing on top of that work. Happy to discuss if there's a use case the generic shape wouldn't cover.

---

aparajon:
Implemented in https://github.com/block/schemabot/pull/386 and https://github.com/block/schemabot/pull/383. We can iterate on the API shape, but closing this for now

PR #121, opened 2026-05-19
feat(scheduler): add durable apply dispatch and retries

aparajon:
## Summary

This makes apply execution durable and scheduler-owned: `/api/apply` persists apply/task records, wakes scheduler workers, and returns once work is queued. Workers claim applies from storage and resume local or remote Tern work, letting SchemaBot recover from request cancellation, server crashes, and retryable engine failures without restarting completed work.

- Add `failed_retryable` apply/task states, retry-budget expiration, progress/API state mapping, and scheduler retry behavior
- Extend local and gRPC Tern clients so `ResumeApply` handles queued dispatch, stale recovery, and retryable re-dispatch
- Enforce one active apply per database/type/environment with MySQL named locks and document scheduler claims, workers, and retry semantics
- Add integration/e2e coverage for queued dispatch, crash recovery, retryable Spirit failures, multi-worker scheduling, and split CI execution modes

🤖 Generated by Codex.

PR #127, opened 2026-05-20
fix(storage): use named apply target locks

aparajon:
## Summary

- Serialize active apply create/update with a per-target MySQL named lock held on a dedicated connection
- Remove hot-path dependency on lazy `apply_target_locks` row creation before checking active applies
- Add regression coverage for target-lock table contention and apply concurrency behavior
- Update architecture docs to describe named apply-target locks

🤖 Generated by Codex on behalf of @aparajon.


aparajon:
Will be handled in https://github.com/block/schemabot/pull/121

Issue #133, opened 2026-05-21
TOCTOU: apply -y gate evaluates against discovery SHA, never re-checks against fresh HEAD before apply


Kiran01bm:
## Summary

In the `apply -y` (auto-confirm) flow the checks gate
(`enforcePassingChecks`) runs against the **discovery** PR HEAD SHA,
and is **never** re-run on the fresh GitHub HEAD before the apply
executes. A check that turned red (or a new required check that
appeared) between discovery and apply will not be observed by the
gate; the apply proceeds against stale check state.

`apply-confirm` is not affected: PR #114 already placed
`FetchPullRequestNoCache` immediately before `enforcePassingChecks`
in `handleApplyConfirmCommand` so the gate there runs on the
current HEAD.

This is a sibling concern to the [stale-schema-files] issue
(`schemaResult.SchemaFiles` is also still the discovery snapshot at
`executeApply` time) but it is structurally distinct: that one is
about the **bytes** being applied, this one is about the **gate
that authorises the apply**. They share a root cause —
discovery-time data being reused at execution time — but they have
different fix shapes and need separate handling.

## Gate position differs between flows

| Step | `apply-confirm` (`handleApplyConfirmCommand`) | `apply -y` (`handleApplyCommand` + AutoConfirm branch) |
|---|---|---|
| 1 | Discovery — `CreateSchemaRequestFromPR` → discovery SHA = `abc`, files@`abc` | Discovery — `CreateSchemaRequestFromPR` → discovery SHA = `abc`, files@`abc` |
| 2 | `enforceReviewGate` (CODEOWNERS, no SHA dependency) | `enforceReviewGate` (CODEOWNERS, no SHA dependency) |
| 3 | **`FetchPullRequestNoCache` → fresh HEAD `def`** | — |
| 4 | **`enforcePassingChecks(def)`** ✅ fresh SHA | **`enforcePassingChecks(prInfo.HeadSHA = abc)`** ⚠️ discovery SHA (cached fetch) |
| 5 | Lock ownership check (already acquired) | Lock check + acquire |
| 6 | — | Plan computed |
| 7 | — | `if result.AutoConfirm` branch entered |
| 8 | — | `FetchPullRequestNoCache` → fresh HEAD `def` |
| 9 | — | `shaVerified := check.HeadSHA == prInfo.HeadSHA` (a SHA-equality test, **not** a re-run of `enforcePassingChecks` against `def`) |
| 10 | `executeApply(...)` | `executeApply(...)` |

[...]

PR #134, opened 2026-05-21
fix(github): reject apply/plan when PR head advances after discovery

### LINKED ISSUE #132: TOCTOU: Stale schema files at apply/plan when HEAD moves between discovery and execution
## Summary

When a new commit lands on a PR branch *between* the discovery step
(`CreateSchemaRequestFromPR`, which loads schema files at the
current HEAD) and the execution step, SchemaBot can run an apply
against the **previous commit's schema files** while the checks
gate validates the **new commit's HEAD**. The same staleness window
exists in the `plan` command: the rendered plan comment is computed
against the discovery-time files even if HEAD has already moved.

`apply-confirm` is fully exposed. `apply -y` is partially mitigated
by an unrelated check (`shaVerified` compares a *prior* stored
check record's SHA against the current HEAD) but the mitigation
holds only by coincidence — it's wired to the wrong invariant and
can let stale schema through in legitimate sequences. `plan` is
fully exposed and silently posts a comment rendered against the
old files; that comment then anchors the user's `apply-confirm`
decision, so the bad artifact propagates forward.

## Pre-#114 vs Post-#114 state

| Path | Pre-#114 | Post-#114 (today) | After this fix |
|---|---|---|---|
| `apply-confirm` — gate runs on which SHA? | discovery (stale) | **fresh ✅** | fresh |
| `apply-confirm` — apply runs on which schema files? | discovery (stale) | **discovery (still stale) ❌** | **rejected if HEAD moved ✅** |
| `apply -y` — SHA verification | discovery (stale) | fresh-vs-stored, can accidentally pass with stale schema files ⚠️ | **rejected if HEAD ≠ discovery ✅** |
| `plan` — comment rendered against which SHA? | discovery (stale; no NoCache fetch) | discovery (still stale; no NoCache fetch) | **rejected if HEAD moved ✅** |
| User-visible response on stale-schema | apply silently runs against old files; plan comment silently rendered against old files | apply silently runs against old files; plan comment silently rendered against old files | clear reject comment (+ lock released for `apply-confirm`) + retry message |

PR #114 introduced `FetchPullRequestNoCache` and used it to make the
checks **gate** read fresh GitHub HEAD. That fixed gate freshness
but did not address the schema **files**, which are loaded once at
discovery time and reused unchanged by `executeApply` regardless of
how much wall time or how many commits have intervened.

## Flow 1 — `apply-confirm`

`pkg/webhook/apply_handlers.go` — `handleApplyConfirmCommand`

```
T0   CreateSchemaRequestFromPR(...)
     ├─ FetchPullRequest         ─▶ HEAD=abc
     ├─ FetchSchemaFilesOptimized @ abc
     └─ schemaResult.HeadSHA       = abc      ← single source of truth!

⇣ new commit lands ⇣                          HEAD = def

T2   FetchPullRequestNoCache     ─▶ HEAD=def  (correctly fresh)
T3   enforcePassingChecks(headSHA=def, ...)    ✅ validates def's checks
T4   executeApply(ctx, ..., schemaResult, ...)
     └─ SchemaFiles: schemaResult.SchemaFiles ← files@abc !
        ❌ stale-schema apply: gate at def, apply at abc
```

Nothing between `CreateSchemaRequestFromPR` and `executeApply`
compares `schemaResult.HeadSHA` against the fresh HEAD.

## Flow 2 — `apply -y` (auto-confirm)

`pkg/webhook/apply_handlers.go` — `handleApplyCommand` (auto-confirm branch)

Same root cause. The current `shaVerified` check compares two SHAs
that don't include `schemaResult.HeadSHA`:

```go
check, _   := h.service.Storage().Checks().Get(...)
prInfo, _  := client.FetchPullRequestNoCache(ctx, repo, pr)
shaVerified := check != nil && prInfo != nil &&
               check.HeadSHA == prInfo.HeadSHA
```

`check.HeadSHA` is from a **prior** stored apply-plan check
record — not "the SHA we just loaded schema files at in this
delivery". The race fires when `check.HeadSHA == prInfo.HeadSHA`
happens to hold while `schemaResult.HeadSHA` differs:

```
T0.0 handleApplyCommand starts
T0.1 ctx = WithPRInfoCache(ctx)                              (PR #114)
T0.2 CreateSchemaRequestFromPR(...)
     ├─ FetchPullRequest    cache MISS ─▶ HEAD=abc
     ├─ FetchSchemaFilesOptimized @ abc
     └─ schemaResult.HeadSHA = abc                ← discovery snapshot

T0.5 lock acquired
T0.6 plan computed (against schemaResult.SchemaFiles = files@abc)

⇣ new commit lands on the PR branch ⇣                       HEAD = def

T1.0 AutoConfirm branch entered
T1.1 check, _ := Storage().Checks().Get(...)
     └─ from a prior `apply -y` at SHA def:   check.HeadSHA = def
T1.2 prInfo, _ := FetchPullRequestNoCache(...)   ─▶ HEAD = def  (fresh)
T1.3 shaVerified := (check.HeadSHA == prInfo.HeadSHA)
                  = (def == def)
                  = true                          ⚠️ wrong invariant
                  ─ does NOT compare schemaResult.HeadSHA (abc)
                    against prInfo.HeadSHA (def)
T1.4 storedPlan loaded; executeApply(..., schemaResult, ...)
     └─ SchemaFiles: schemaResult.SchemaFiles = files@abc
        ❌ stale-schema apply: shaVerified passed but on the wrong pair
```

## Flow 3 — `plan`

`pkg/webhook/plan.go` — `handlePlanCommand`

Unlike `apply-confirm`, the plan path has **no** `FetchPullRequestNoCache`
call at all today. Discovery loads files at `abc`, and `ExecutePlan` /
the rendered comment is computed against those exact files — even if
HEAD has already moved to `def` by the time the comment is posted.

```
T0   CreateSchemaRequestFromPR(...)
     ├─ FetchPullRequest         ─▶ HEAD=abc
     ├─ FetchSchemaFilesOptimized @ abc
     └─ schemaResult.HeadSHA       = abc      ← single source of truth!

⇣ new commit lands ⇣                          HEAD = def
     (no NoCache fetch here — gap is silent)

T2   ExecutePlan(ctx, ..., schemaResult, ...)
     └─ SchemaFiles: schemaResult.SchemaFiles ← files@abc !
        ❌ stale-schema plan: comment rendered against abc
           while branch is at def
T3   PR comment posted titled / anchored on abc-derived DDL
     └─ user reacts with `apply-confirm` against this comment;
        the stale artifact has now propagated forward into the
        apply flow before any guard ever runs.
```

`plan` does not mutate the database, so the immediate blast
radius is smaller — but the comment misleads the human reviewer,
who then issues `apply-confirm` against an out-of-date plan.
Closing this window at `plan` stops the bad artifact at its
source rather than catching it later in the chain.

## Invariants violated

- **Started applies remain authoritative.** Apply runs files@abc
  while the truth on the PR branch is at def; the just-pushed def
  schema is silently ignored.
- **TOCTOU review.** The NoCache fetch is run for gates but not
  for files, so the TOCTOU check protects the gate and not the
  artifact the gate exists to protect.
- **Storage / API uncertainty must never be converted into a
  passing check.** A passing gate at def is implicitly used to
  justify executing files from abc — gate passes for the wrong
  artifact.

## The missing invariant

> Before any `executeApply` **or** `ExecutePlan`,
> `schemaResult.HeadSHA` MUST equal the fresh GitHub HEAD obtained
> via `FetchPullRequestNoCache`.

If it doesn't, the artifact in front of the user — whether the
DDL about to be executed, the plan comment they reacted to, or
the implicit auto-plan in `-y` — is for a different commit than
the one whose code is currently on the PR branch. The only safe
response is to stop, tell the user, and require them to re-issue
the command so discovery picks up the new HEAD.

## Related issues

- Sibling: [stale-gate] — `apply -y` runs `enforcePassingChecks`
  against the *discovery* SHA and never re-evaluates against the
  fresh HEAD before `executeApply`. Shares the same root cause
  (discovery-time data reused at execution time) but a different
  window — schema-file bytes vs check-state evaluation.

## References

- PR #114 (introduced `FetchPullRequestNoCache` and fixed the
  gate-freshness half of this race).
- PR #112 (related: singleflight coalescer for check status —
  different race, different fix).
- Sibling issue: [stale-gate] — gate-evaluation freshness on
  `apply -y`.

[stale-gate]: https://github.com/block/schemabot/issues/133

---

Kiran01bm:
## What and Why ?
Fixes - https://github.com/block/schemabot/issues/132

Discovery (`CreateSchemaRequestFromPR`) loads schema files at the cached PR HEAD. If a new commit lands before execution, the previous behaviour silently ran `executeApply` or rendered a plan comment against stale bytes — DDL derived from a commit the branch was no longer on.

`assertSchemaStillCurrent` compares the discovery `HeadSHA` against a fresh `FetchPullRequestNoCache` result immediately before `plan`, `apply` (auto-confirm `-y`), and `apply-confirm`. On mismatch: log, increment `schemabot.apply.rejected_stale_schema`, post a rejection comment, and (for the apply paths) release the per-target lock so the user can re-run cleanly.

The previous `shaVerified` check in the auto-confirm branch compared a stored prior-render `HeadSHA` against the fresh fetch — a coincidence-check across deliveries that could pass while `schemaResult.HeadSHA` differed. Replaced by the helper.

Closes #132.

The sibling `apply -y` gate-freshness window (#133) is intentionally out of scope.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -100,30 +100,31 @@ func RecordApplyDuration(ctx context.Context, duration time.Duration, database, 
 	)
 }
 
-// knownStaleSchemaActions limits metric cardinality on the
-// schemabot.apply.rejected_stale_schema counter to the three handlers that
+// knownSchemaFreshnessActions limits metric cardinality on the
+// schemabot.schema_freshness.rejected counter to the three handlers that
 // load a schema snapshot at discovery and reuse it at execution.
-var knownStaleSchemaActions = map[string]bool{
+var knownSchemaFreshnessActions = map[string]bool{
 	"plan":          true,
 	"apply":         true,
 	"apply_confirm": true,
 }
 
-// RecordApplyRejectedStaleSchema increments the counter for plan/apply/apply-confirm
+// RecordSchemaFreshnessRejected increments the counter for plan/apply/apply-confirm
 // commands rejected because the PR branch HEAD advanced after discovery loaded the
-// schema files. A spike indicates aggressive force-pushing, webhook replay, or a
-// regression in the schema-freshness guard.
-func RecordApplyRejectedStaleSchema(ctx context.Context, action string) {
-	if !knownStaleSchemaActions[action] {
+// schema files. The metric name is action-neutral because the same rejection fires
+// for read-only plan as well as mutating apply paths. A spike indicates aggressive
+// force-pushing, webhook replay, or a regression in the schema-freshness guard.
+func RecordSchemaFreshnessRejected(ctx context.Context, action string) {
+	if !knownSchemaFreshnessActions[action] {
 		action = "unknown"
 	}
 	meter := otel.Meter(meterName)
-	counter, err := meter.Int64Counter("schemabot.apply.rejected_stale_schema.total",
+	counter, err := meter.Int64Counter("schemabot.schema_freshness.rejected.total",
 		otelmetric.WithDescription("Plan/apply/apply-confirm rejected because PR HEAD advanced after discovery loaded schema files"),
 		otelmetric.WithUnit("{rejection}"),
 	)
 	if err != nil {
-		slog.Warn("failed to create apply rejected stale schema counter", "error", err)
+		slog.Warn("failed to create schema freshness rejected counter", "error", err)
 		return
 	}
 	counter.Add(ctx, 1,
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -100,6 +100,39 @@ func RecordApplyDuration(ctx context.Context, duration time.Duration, database, 
 	)
 }
 
+// knownStaleSchemaActions limits metric cardinality on the
+// schemabot.apply.rejected_stale_schema counter to the three handlers that
+// load a schema snapshot at discovery and reuse it at execution.
+var knownStaleSchemaActions = map[string]bool{
+	"plan":          true,
+	"apply":         true,
+	"apply_confirm": true,
+}
+
+// RecordApplyRejectedStaleSchema increments the counter for plan/apply/apply-confirm
+// commands rejected because the PR branch HEAD advanced after discovery loaded the
+// schema files. A spike indicates aggressive force-pushing, webhook replay, or a
+// regression in the schema-freshness guard.
+func RecordApplyRejectedStaleSchema(ctx context.Context, action string) {
+	if !knownStaleSchemaActions[action] {
+		action = "unknown"
+	}
+	meter := otel.Meter(meterName)
+	counter, err := meter.Int64Counter("schemabot.apply.rejected_stale_schema.total",
+		otelmetric.WithDescription("Plan/apply/apply-confirm rejected because PR HEAD advanced after discovery loaded schema files"),
+		otelmetric.WithUnit("{rejection}"),
+	)
+	if err != nil {
+		slog.Warn("failed to create apply rejected stale schema counter", "error", err)
+		return
+	}
+	counter.Add(ctx, 1,
+		otelmetric.WithAttributes(
+			attribute.String("action", action),
+		),
+	)
+}
+
 // knownCheckOwnershipOperations limits metric cardinality to expected check
 // ownership miss paths.
 var knownCheckOwnershipOperations = map[string]bool{
--- a/pkg/webhook/apply_e2e_test.go
+++ b/pkg/webhook/apply_e2e_test.go
@@ -1178,26 +1178,14 @@ func TestE2EApplyAutoConfirmExecutes(t *testing.T) {
 		"expected apply summary comment")
 }
 
-// TestE2EApplyAutoConfirmDowngradesOnSHAMismatch verifies that -y downgrades to
-// manual confirmation when the HEAD SHA doesn't match the stored check record.
-func TestE2EApplyAutoConfirmDowngradesOnSHAMismatch(t *testing.T) {
-	dbName := "webhook_auto_confirm_sha"
+// TestE2EApplyAutoConfirmRejectsWhenHEADAdvanced verifies that `apply -y`
+// rejects (does NOT apply, releases the lock) when the PR HEAD advances
+// between discovery (cached FetchPullRequest) and the fresh
+// FetchPullRequestNoCache fetch inside the auto-confirm branch. The user
+// must re-run the command so discovery picks up the new HEAD.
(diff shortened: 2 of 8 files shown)


aparajon:
🤖 Generated by Codex on behalf of @aparajon.

I think this still misses one stale-plan case for manual `schemabot apply`:

1. `schemabot apply` discovers schema at `abc123`.
2. PR head advances to `def456` before the confirmation plan is posted.
3. The handler can still post the confirmation plan for `abc123`.
4. Later `schemabot apply-confirm` runs a new discovery at `def456`, then `FetchPullRequestNoCache` also sees `def456`, so the new `schemaResult.HeadSHA == current head` check passes.

So the confirm-time check catches "head advanced during this apply-confirm handler", but it does not prove that the pending confirmation plan the user reviewed was generated for the current PR head.

I think we need either:
- fresh-check PR head before posting the manual apply confirmation plan, and reject/retry if it moved; and/or
- persist/compare the plan/check/lock SHA during `apply-confirm` so confirm rejects if the pending plan SHA != current PR head.

Separate note: the stale confirm rejection currently calls `ForceRelease` before the lock ownership check, which looks like it could release another PR's lock for the same database/type. That should probably verify ownership before any force-release.

pkg/metrics/metrics.go
```diff
@@ -100,6 +100,39 @@ func RecordApplyDuration(ctx context.Context, duration time.Duration, database,
...
+// regression in the schema-freshness guard.
+func RecordApplyRejectedStaleSchema(ctx context.Context, action string) {
+	if !knownStaleSchemaActions[action] {
+		action = "unknown"
+	}
+	meter := otel.Meter(meterName)
+	counter, err := meter.Int64Counter("schemabot.apply.rejected_stale_schema.total",
+		otelmetric.WithDescription("Plan/apply/apply-confirm rejected because PR HEAD advanced after discovery loaded schema files"),
+		otelmetric.WithUnit("{rejection}"),
+	)
```
aparajon: Yeah I think I agree with this comment, perhaps we just use  `schemabot.command.rejected_stale_schema.total` without `.apply`

Kiran01bm:
> I think we need either:
> 
> * fresh-check PR head before posting the manual apply confirmation plan, and reject/retry if it moved; and/or
> * persist/compare the plan/check/lock SHA during `apply-confirm` so confirm rejects if the pending plan SHA != current PR head.

good point - ive done the fresh-check in this pr as this is within the delivery window - and for 2nd one i've filed this issue - https://github.com/block/schemabot/issues/142 as we need both..

aparajon (approved):

PR #143, opened 2026-05-22
fix(github): reject apply-confirm when the stored plan was rendered against a stale PR HEAD

### LINKED ISSUE #142: apply-confirm can execute DDL for a commit the user never reviewed (cross-delivery HEAD race)
## Summary

PR #134 added `assertSchemaStillCurrent` guards across `plan`, manual `apply`, `apply -y`, and `apply-confirm`. Each guard closes a **within-delivery** TOCTOU window — the PR HEAD advancing while a single webhook handler is running. None of them close the **cross-delivery** window: HEAD advancing *between* the plan being posted and the user clicking `apply-confirm`. At confirm time the user has reviewed and approved DDL for a commit that is no longer the branch HEAD.

## The window this issue closes

1. `schemabot apply` discovers schema at `abc123`, posts a confirmation plan for `abc123`.
2. PR HEAD advances to `def456` (e.g. someone pushes a commit) — the manual-apply guard in PR #134 cannot fire because that handler has already returned.
3. User reads the `abc123` plan and comments `schemabot apply-confirm`.
4. `apply-confirm` runs a fresh discovery at `def456`. The confirm-time freshness check compares `schemaResult.HeadSHA` (`def456`) against `FetchPullRequestNoCache` (`def456`) — they match, so the check passes.
5. The apply proceeds against `def456` even though the user only ever saw / approved the `abc123` plan.

The PR #134 guards only catch races *within one webhook delivery*; this race spans **across** deliveries (plan delivery → user review → confirm delivery).

## Proposed fix

- Persist `schemaResult.HeadSHA` (the SHA the plan was rendered against) at the moment the confirmation plan is posted.
  - Storage shape: add `plan_head_sha` to either the `plans` row or the `checks` row created for the `apply_plan` check; the `plans` row is the most natural home since it already holds the rendered `schema_files`.
- At `apply-confirm`, compare the current PR HEAD (from `FetchPullRequestNoCache`) against the **stored plan's SHA**, not against the confirm-time discovery SHA.
- On mismatch: reject with a "plan you confirmed is now stale, re-run `schemabot apply -e <env>`" comment and release the lock (owner-scoped `Release`, per the precedent set in PR #134 — not `ForceRelease`).
- Add an integration test that posts a plan at SHA A, advances the branch to SHA B, then runs `apply-confirm` and asserts the cross-delivery guard rejects.

## Relation to PR #134

This is intentionally separate from PR #134 — `assertSchemaStillCurrent` closes *within-delivery* TOCTOU windows at every call site; this issue closes the *cross-delivery* window that no in-handler guard can see. Both are required for the full safety story.

## Source

Identified by @aparajon in [PR #134 review comment](https://github.com/block/schemabot/pull/134#issuecomment-4507859337).

---

*This issue was generated by Amp (claude-sonnet-4-5).*

Kiran01bm:
## What and Why ?

Closes #142.

PR #134 added in-handler guards that reject when the PR HEAD advances *during*
a single webhook delivery. They cannot see the cross-delivery race: HEAD
advancing between the confirmation plan being posted and the user clicking
`apply-confirm`. At confirm time both the new discovery and the fresh fetch see
the same new SHA, so #134's within-delivery comparison trivially passes — but
the plan the user reviewed and approved was rendered for the older commit.

This PR closes that gap by persisting `schemaResult.HeadSHA` on the stored
`plans` row (new `head_sha` column, plumbed through `apitypes.PlanRequest`,
`api.PlanRequest`, and the Tern `PlanRequest` proto so both GRPCClient and
LocalClient paths populate it). At `apply-confirm` time, after the existing
ownership gate, the handler loads the latest stored plan for the target and
compares its `HeadSHA` against `FetchPullRequestNoCache`. On mismatch it posts
a rejection comment that names both SHAs and tells the user to re-run
`schemabot apply -e <env>` to get a fresh plan, then releases the lock.

Lock release uses `ForceRelease` (not owner-scoped `Release`) because the
cross-delivery check fires *after* the ownership gate — at that point we have
verified this PR owns the lock. The behavior is pinned by
`TestE2EApplyConfirmStalePlanPreservesOtherPRLock`, the cross-delivery sibling
of #134's `TestE2EApplyConfirmStaleSchemaPreservesOtherPRLock`: if a future
refactor moves the cross-delivery check above the ownership gate, that test
will fail and force the author to either keep the ordering or switch to
owner-scoped Release at the same time.

A command-neutral metric `schemabot.command.rejected_stale_plan.total` is
emitted on rejection. Distinct from `schemabot.schema_freshness.rejected.total`
because operator triage differs: spikes here indicate humans pushing during PR
review (across deliveries), not webhook races (within delivery).

### Skip vs fail-closed for legacy `head_sha` rows

Plans created before this column existed have `head_sha = ''`. The helper
treats empty as *skip* (the invariant cannot be evaluated) rather than
*fail closed*. Reviewer call: we considered fail-closed and could go either
way since SchemaBot is not yet live in production — a fail-closed default
would let us assert "every plan is verifiable or rejected", at the cost of
rejecting any in-flight `apply` issued before the new server image was
deployed. We erred on the side of operator friction (skip) so a deploy never
makes a pending `apply-confirm` unapproveable. Happy to flip it to fail-closed
if reviewers prefer; it is a one-line change and a couple of test tweaks.


pkg/webhook/plan_freshness.go
```diff
@@ -0,0 +1,130 @@
...
+	pr int,
+	environment, database, databaseType string,
+) (*storage.Plan, error) {
+	plans, err := h.service.Storage().Plans().GetByPR(ctx, repo, pr)
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range plans {
+		if p.Environment == environment && p.Database == database && p.DatabaseType == databaseType {
+			return p, nil
```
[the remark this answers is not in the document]
Kiran01bm: :-) amazing pick-up codex! well done.. perhaps i'll add a `id DESC` as a tiebreaker

pkg/webhook/apply_handlers.go
```diff
@@ -423,6 +424,35 @@ func (h *Handler) handleApplyConfirmCommand(repo string, pr int, environment, da
...
+		h.postComment(repo, pr, installationID, templates.RenderGenericError(templates.SchemaErrorData{
+			RequestedBy: requestedBy,
+			Environment: environment,
+			CommandName: action.ApplyConfirm,
+			ErrorDetail: "Failed to load stored plan: " + planLoadErr.Error(),
+		}))
+		return
+	}
+	if rejected := h.assertPlanStillCurrent(ctx, repo, pr, installationID, storedPlan, confirmPRInfo.HeadSHA, environment, requestedBy); rejected {
+		if relErr := h.service.Storage().Locks().ForceRelease(ctx, database, dbType); relErr != nil {
```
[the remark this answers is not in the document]
Kiran01bm: fair point - i'll mirror the stale-schema rejection pattern and switch to the owner-scoped release.. ty!

pkg/webhook/plan_freshness.go
```diff
@@ -0,0 +1,130 @@
...
+// the first row matching the target filters is the plan the user is about to
+// confirm. Used by handleApplyConfirmCommand to load the plan for the
+// cross-delivery freshness check.
+func (h *Handler) latestPlanForTarget(
+	ctx context.Context,
+	repo string,
+	pr int,
+	environment, database, databaseType string,
+) (*storage.Plan, error) {
+	plans, err := h.service.Storage().Plans().GetByPR(ctx, repo, pr)
```
aparajon: 🤖 Generated by Codex on behalf of @aparajon.

I think this lookup can still let a stale pending apply plan through. `latestPlanForTarget` picks the newest stored plan for `repo + PR + env + database + type`, but `plans` is shared by `schemabot apply` confirmation plans and plain `schemabot plan` results. A newer plain plan can therefore supersede the pending apply plan for this lookup.

Real-world sequence:

1. Someone comments `schemabot apply -e staging`.
2. SchemaBot renders an apply confirmation plan from PR head `abc123`, posts it, and acquires the apply lock.
3. Before anyone confirms, someone pushes another commit, so PR head becomes `def456`.
4. A normal plan gets generated at `def456` (for example, someone comments `schemabot plan -e staging`, an auto-plan webhook runs after the push, or someone reruns plan while reviewing the new commit).
5. The reviewer scrolls back to the older apply confirmation comment from step 2 and runs `schemabot apply-confirm -e staging`.
6. This PR asks for the latest stored plan for the target, finds the newer plain plan from step 4 at `def456`, and the freshness check passes.
7. SchemaBot proceeds even though the human confirmed a plan rendered for `abc123`.

So the issue is not that SchemaBot necessarily applies old bytes directly; `apply-confirm` re-plans from current PR files before executing. The safety gap is that the human approval is no longer tied to the exact plan they reviewed. A newer plain plan can accidentally make an old confirmation comment look current.

I proved this locally against PR head `6fb9e44`: seed a pending apply plan at `abc123`, acquire the apply lock, run a normal `schemabot plan -e staging` at `newsha456`, then run `schemabot apply-confirm -e staging`. Expected rejection because the confirmation plan the user reviewed was still the `abc123` apply plan. Actual result: `apply-confirm` posted `Schema Change In Progress` and created an apply, because the guard loaded the newer plain plan at `newsha456`.

I think the fix needs to tie confirmation to the apply artifact itself, e.g. store the pending `plan_id` or `plan_head_sha` on the lock when `schemabot apply` posts the confirmation plan, then compare that exact value during `apply-confirm`.

<details>
<summary>Failing proof test</summary>

```go
func TestE2EApplyConfirmRejectsWhenPlainPlanSupersedesPendingApplyPlan(t *testing.T) {
	dbName := "webhook_confirm_plain_plan_supersedes_apply_plan"
	svc := setupE2EService(t, dbName)

	seedCheck(t, svc, dbName, "staging", "action_required")

	require.NoError(t, svc.Storage().Locks().Acquire(t.Context(), &storage.Lock{
		DatabaseName: dbName,
		DatabaseType: "mysql",
		Repository:   "octocat/hello-world",
		PullRequest:  1,
		Owner:        "octocat/hello-world#1",
	}))
	t.Cleanup(func() {
		_ = svc.Storage().Locks().ForceRelease(context.WithoutCancel(t.Context()), dbName, "mysql")
	})

	_, err := svc.Storage().Plans().Create(t.Context(), &storage.Plan{
		PlanIdentifier: "apply_plan_stale_abc123",
		Database:       dbName,
		DatabaseType:   "mysql",
		Deployment:     dbName,
		Target:         dbName,
		Repository:     "octocat/hello-world",
		PullRequest:    1,
		Environment:    "staging",
		HeadSHA:        "abc123",
		CreatedAt:      time.Now().Add(-time.Minute),
	})
	require.NoError(t, err)

	mux := http.NewServeMux()
	server := httptest.NewServer(mux)
	t.Cleanup(server.Close)

	client := gh.NewClient(nil)
	client.BaseURL, _ = url.Parse(server.URL + "/")

	schemabotConfig := fmt.Sprintf("database: %s\ntype: mysql\n", dbName)
	schemaFiles := map[string]string{
		"users.sql": "CREATE TABLE `users` (\n  `id` bigint unsigned NOT NULL AUTO_INCREMENT,\n  `name` varchar(255) NOT NULL,\n  PRIMARY KEY (`id`)\n) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;",
	}

	result := setupFakeGitHubForPlan(t, mux, schemaFiles, schemabotConfig, dbName)
	result.HeadSHAs = []string{"newsha456"}
	h := newE2EHandler(t, svc, client)

	planReq := buildWebhookRequest(t, webhookPayloadOpts{
		comment: "schemabot plan -e staging",
		isPR:    true,
	}, nil)
	planRR := httptest.NewRecorder()
	h.ServeHTTP(planRR, planReq)
	require.Equal(t, http.StatusOK, planRR.Code)

	select {
	case body := <-result.comments:
		require.Contains(t, body, "Schema Change Plan")
		require.NotContains(t, body, "Lock acquired by")
	case <-time.After(30 * time.Second):
		t.Fatal("timed out waiting for plain plan comment")
	}

	confirmReq := buildWebhookRequest(t, webhookPayloadOpts{
		comment: "schemabot apply-confirm -e staging",
		isPR:    true,
	}, nil)
	confirmRR := httptest.NewRecorder()
	h.ServeHTTP(confirmRR, confirmReq)
	require.Equal(t, http.StatusOK, confirmRR.Code)

	select {
	case body := <-result.comments:
		assert.Contains(t, body, "Rejected")
		assert.Contains(t, body, "the plan you confirmed is stale")
		assert.Contains(t, body, "abc123")
		assert.Contains(t, body, "newsha456")
		assert.NotContains(t, body, "Schema Change In Progress")
		assert.NotContains(t, body, "Schema Change Applied")
	case <-time.After(30 * time.Second):
		t.Fatal("timed out waiting for apply-confirm response")
	}

	applies, err := svc.Storage().Applies().GetByPR(t.Context(), "octocat/hello-world", 1)
	require.NoError(t, err)
	for _, a := range applies {
		assert.NotEqual(t, dbName, a.Database, "stale pending apply plan must not be executable after a newer plain plan")
	}
}
```

</details>

Kiran01bm: good catch - this is now fixed - ptal..

Kiran01bm: actually did not notice a pending conflict.. resolving it now

aparajon (approved):

PR #169, opened 2026-05-27
refactor(state): add shared apply state metadata registry

Kiran01bm:
## What and Why ?
Adds a single source of truth in `pkg/state` for canonical Apply state metadata — human label, terminal classification, and PlanetScale branch-setup classification — replacing three separate switches that had to be kept in sync.

- `pkg/state/metadata.go`: `ApplyStateInfo {Label, Terminal, BranchSetup}` and `applyMetadata` covering every `Apply.*` constant; `LookupApply` and `Label` helpers.
- `pkg/state/apply.go`: `IsTerminalApplyState` and `IsBranchSetupPhase` consult the registry instead of open-coded switches.
- `pkg/cmd/templates/progress.go`: drop the local `StateLabel` switch; CLI call sites use `state.Label` directly.
- `pkg/state/metadata_test.go`: exhaustiveness test plus pinned-set tests for `Terminal` and `BranchSetup` so accidental classification drift fails loudly.

### Benefits

- **One place to add a new Apply state.** A single registry entry sets label, terminal, and branch-setup behavior at once; reviewers can no longer forget one of three switches.
- **Exhaustiveness is enforced.** A missing registry entry fails `TestApplyMetadata_CoversAllStates` instead of silently rendering the raw state string in the CLI or quietly returning `false` from gating predicates.
- **Terminal and branch-setup classifications are pinned.** Pinned-set tests turn changes to these sets — which affect scheduler claiming, reconciliation, merge gating, and per-table progress display — into deliberate, reviewable diffs.
- **Removes duplicated logic across packages.** CLI labels and structural classifications were previously maintained in three files; they now derive from one map.
- **Safe default for unknown states.** `Label` falls back to the raw input rather than panicking, matching prior CLI behavior.
- **Registry stays presentation-neutral.** Surface-specific copy (Title Case PR headers, ANSI colors, emojis) remains in each renderer so the registry does not overfit to one surface.

### DIFF
--- a/pkg/state/metadata.go
+++ b/pkg/state/metadata.go
@@ -3,7 +3,7 @@ package state
 // ApplyStateInfo holds presentation-neutral metadata for a canonical Apply state.
 //
 // Renderers (CLI, TUI, PR comments) reach for Label when they need a short
-// human-readable name. Control-plane code reaches for Terminal and BranchSetup
+// human-readable name. Control-plane code reaches for Terminal and SetupPhase
 // to make scheduling and gating decisions. Centralizing this metadata avoids
 // the drift that occurs when each consumer maintains its own switch.
 //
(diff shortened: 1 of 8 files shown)


pkg/state/metadata.go
```diff
@@ -0,0 +1,119 @@
...
+	// FailedRetryable is NOT terminal — the scheduler may retry it.
+	// Stopped IS terminal at the apply level (operators must explicitly resume).
+	Terminal bool
+
+	// BranchSetup is true for PlanetScale branch-lifecycle phases where
+	// per-table progress is not yet meaningful (all tables are queued).
+	// Used by the CLI and TUI to suppress the table list during setup.
+	// Pending and WaitingForDeploy are included because the deploy hasn't
+	// started yet.
+	BranchSetup bool
```
aparajon: Can we rename this to `SetupPhase`? So we can make this more engine agnostic in the future as we onboard more engines

Kiran01bm: sure, the change is now in..

aparajon (approved):

PR #173, opened 2026-05-27
refactor(proto): scope control requests by apply id

aparajon:
Remove database/type fields from apply-scoped control RPC requests so cutover, stop, start, volume, revert, and skip-revert identify work by apply_id plus environment. The HTTP API already derives database metadata from the stored apply row, and the Tern server now rejects control RPCs that omit apply_id.

🤖 Generated with Codex.


pkg/proto/tern.proto
```diff
@@ -467,15 +455,12 @@ message StopResponse {
...
-  // Database type: "vitess" or "mysql". Used by stateless clients.
-  string type = 2;
+  // Apply ID for the schema change to start.
+  string apply_id = 1;
   // Environment: "staging" or "production".
-  string environment = 3;
-  // Scope start to a specific apply ID.
-  string apply_id = 4;
+  string environment = 2;
 }
```
morgo, line 479: Are you concerned about reordering proto fields? Anything inflight breaks, but there shouldn't be too much of that.

morgo (approved):

pkg/proto/tern.proto
aparajon: I'll publicly say we're still in ship fast and break things mode 😅 . Once we're fully live in production at Block, we'll remove experimental status and have a proper policy around breaking changes

PR #230, opened 2026-06-04
fix(scheduler): guard stale worker writes

aparajon:
## Why
Workers can be recovered after an apply heartbeat becomes stale, but an old worker or observer may still be running locally. Control-plane state should only be writable by the claim that currently owns the apply, so stale workers fail closed instead of publishing outdated state.

## What
- Add a rotating apply lease token when an apply is claimed
- Guard apply, task, log, heartbeat, control-request, and observer side effects by the active lease
- Cancel local apply work when the heartbeat loop detects lease loss
- Document the split between heartbeat recovery and lease-token write ownership
- Add regression coverage for stale worker writes and stale observer terminal side effects

## Claim lease flow
```text
1. worker A claims the apply

   ╭──────────╮        claim         ╭──────────────────────────╮
   │ worker A │─────────────────────▶│ applies row              │
   │ token A  │                      │ updated_at = fresh       │
   ╰──────────╯                      │ lease_token = token A    │
                                     ╰──────────────────────────╯

2. worker A stops heartbeating long enough for recovery

   ╭──────────╮                      ╭──────────────────────────╮
   │ worker A │                      │ applies row              │
   │ token A  │                      │ updated_at = stale       │
   ╰──────────╯                      │ lease_token = token A    │
                                     ╰──────────────────────────╯

3. worker B claims the stale apply and rotates the token

   ╭──────────╮        claim         ╭──────────────────────────╮
   │ worker B │─────────────────────▶│ applies row              │
   │ token B  │                      │ updated_at = fresh       │
   ╰──────────╯                      │ lease_token = token B    │
                                     ╰──────────────────────────╯

4. worker A wakes up and tries to write with token A

   ╭──────────╮   write WHERE token=A ╭──────────────────────────╮
   │ worker A │──────────────────────▶│ applies row              │
   │ token A  │       rejected        │ lease_token = token B    │
   ╰──────────╯                       ╰──────────────────────────╯
```

Generated with Amp


pkg/tern/local_apply_sequential.go
```diff
@@ -237,6 +238,15 @@ func (c *LocalClient) startApplyHeartbeat(ctx context.Context, apply *storage.Ap
...
 			case <-ticker.C:
 				if err := c.storage.Applies().Heartbeat(hbCtx, apply.ID); err != nil {
+					if errors.Is(err, storage.ErrApplyLeaseLost) {
+						c.logger.Warn("heartbeat failed because apply lease was lost; local worker will stop writing apply state",
+							"apply_id", apply.ApplyIdentifier,
+							"database", apply.Database,
+							"environment", apply.Environment,
+							"error", err)
+						cancel()
+						return
```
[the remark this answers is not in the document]
Kiran01bm: this one seems legit..

pkg/storage/mysqlstore/tasks.go
```diff
@@ -95,19 +95,49 @@ func (s *taskStore) Get(ctx context.Context, taskIdentifier string) (*storage.Ta
...
+		task.StartedAt, task.CompletedAt,
+		task.ID,
+	}
+	leasePredicate := ""
+	if hasLease {
+		leasePredicate = `
+			AND EXISTS (
+				SELECT 1 FROM applies a
+				WHERE a.id = tasks.apply_id AND a.lease_token = ?
+			)`
```
Kiran01bm, line 115: `WHERE lease_token = ?` predicates are on `applies`, `tasks`, `apply_logs`, and `Heartbeat`, but scheduler-owned writes against `apply_operations`, `apply_comments`, `checks`, and the control-request `complete`/`fail` paths are still unconditional at the SQL layer...

perhaps extend (similar to this) same lease-token guard to those so that the check is atomic with the write instead of a separate round-trip.

pkg/webhook/comment_observer.go
```diff
@@ -228,6 +237,27 @@ func (o *CommentObserver) shouldDeferCutover(apply *storage.Apply) bool {
...
+	lease := o.applyLease
+	if !lease.Valid() && apply != nil {
+		lease = apply.Lease()
+	}
+	if !lease.Valid() {
+		return true
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	if err := o.stor.Applies().CheckLease(ctx, lease); err != nil {
```
Kiran01bm, line 265: edit history overwrites - details from my agent - 
```
Apply 42, current lease = TOK_A held by worker A.

t=0ms     OnTerminal fires on A
t=1ms     leaseStillOwnsObserver: CheckLease(TOK_A) → SELECT returns 1 row ✅
t=2ms     returns true
            ────────────────────────────────────────────────────
            window where another worker can rotate the lease
            ────────────────────────────────────────────────────
t=8ms     worker B reclaims → rotates row's lease_token to TOK_B
          (A still holds TOK_A in memory; A's view of ownership is now stale)
t=15ms    A enters editTrackedComment: client.EditIssueComment(...) hits GitHub
            → the "failed: stale worker error" body gets posted to the PR
t=600ms   GitHub responds 200; A then calls IncrementEditCount
            → that storage call is *also* unguarded against the lease, so the
              edit-count row gets bumped under the wrong owner
t=2s      B's OnProgress tries to edit the same comment with the real
          "running" body → PR comment flips back, leaving the operator with
          a confusing edit history
```

Kiran01bm (approved):

PR #50, opened 2026-04-28
feat(scheduler): add retryable failure handling and worker pool semantics

aparajon:
## Summary

Improve scheduler retry recovery so retryable failures can be recovered safely while permanent PlanetScale failures still become terminal.

- Adds `failed_retryable` handling for applies and tasks, with scheduler re-dispatch up to the retry budget before permanent failure
- Adds configurable scheduler worker concurrency with database exclusion and fresh leases for claimed retryable applies
- Keeps engine errors retryable by default for Spirit, while PlanetScale explicitly marks known permanent apply failures non-retryable
- Preserves Vitess resume metadata for PlanetScale deploy requests during recovery
- Covers retry recovery, retry exhaustion, exclusive claims, Vitess resume, LocalScale revert-window behavior, and permanent PlanetScale failure handling in tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #232, opened 2026-06-04
fix(vitess): recover late migration context

aparajon:
> **Status — superseded, do not merge as-is (will be reworked).** The recovery
> logic here is still wanted, but the storage approach is changing. This PR
> persists the migration-context baseline on the `vitess_apply_data` table; that
> table is being removed in favor of carrying PlanetScale engine resume state on
> the operation row (`engine_resume_metadata`). This PR will be reworked to store
> the baseline/recovered contexts in `engine_resume_metadata` instead of a new
> `vitess_apply_data` column, so the fix lands alongside that consolidation rather
> than deepening a table that is going away. The recovery algorithm (baseline
> diff, earliest-timestamp selection, persist from both the worker and API paths)
> is preserved.

---

## Why
SchemaBot can sometimes create/start a PlanetScale deploy request before Vitess exposes the deploy's `migration_context` in `SHOW VITESS_MIGRATIONS`.

When that happens, SchemaBot still knows the deploy request ID, but it cannot show the richer Vitess progress view: per-table/per-shard progress, row counts, ETA, and related migration details. Without a recovery path, progress can stay limited to coarse deploy-request status even after the Vitess rows appear.

## What
- Save the set of Vitess migration contexts that existed before the deploy starts.
- Persist that baseline on `vitess_apply_data` so it survives normal progress polling and restarts.
- If progress polling later sees deploy metadata but no `migration_context`, query `SHOW VITESS_MIGRATIONS` again and recover a context from current rows.
- Select recovered contexts deterministically: ignore baseline contexts, prefer the earliest `requested_timestamp` at or after the deploy request creation time, and avoid guessing when multiple untimed new contexts exist.
- Persist a recovered context from both the active apply worker loop and read/API progress path so recovery survives worker restarts and scheduler owner handoff.
- Store Vitess row timestamps with each baseline context for operator debugging.

## Risk Assessment
Low. This only improves PlanetScale progress enrichment when resume state is missing `migration_context`; deploy-request state remains the source of truth for apply state.

The storage change is additive: a nullable JSON column on `vitess_apply_data`.

Generated with Amp

### DIFF
--- a/pkg/engine/planetscale/progress.go
+++ b/pkg/engine/planetscale/progress.go
@@ -81,7 +81,7 @@ func (e *Engine) Progress(ctx context.Context, req *engine.ProgressRequest) (*en
 		}
 	}
 	if req.ResumeState.MigrationContext == "" && len(meta.ExistingMigrationCtxs) > 0 {
-		migrationContext := e.discoverMigrationContext(ctx, client, req.Database, req.Credentials, meta.ExistingMigrationCtxs)
+		migrationContext := e.discoverMigrationContext(ctx, client, req.Database, req.Credentials, meta.ExistingMigrationCtxs, dr.CreatedAt)
 		if migrationContext != "" {
 			req.ResumeState = &engine.ResumeState{
 				MigrationContext: migrationContext,
@@ -202,13 +202,13 @@ func formatMigrationTimestamp(ts *time.Time) string {
 
 // discoverMigrationContext finds the new migration_context that appeared after
 // deploying by comparing current contexts against the pre-deploy baseline.
-func (e *Engine) discoverMigrationContext(ctx context.Context, client psclient.PSClient, database string, creds *engine.Credentials, existingContexts map[string]MigrationContextTimestamps) string {
+func (e *Engine) discoverMigrationContext(ctx context.Context, client psclient.PSClient, database string, creds *engine.Credentials, existingContexts map[string]MigrationContextTimestamps, deployCreatedAt time.Time) string {
 	if creds.DSN == "" {
 		e.logger.Debug("skipping schema change context discovery, no DSN configured")
 		return ""
 	}
 
-	e.logger.Info("discovering schema change context", "database", database, "baseline_count", len(existingContexts))
+	e.logger.Info("discovering schema change context", "database", database, "baseline_count", len(existingContexts), "deploy_created_at", deployCreatedAt)
 
 	branch := mainBranch(creds)
 	keyspaces, err := client.ListKeyspaces(ctx, &ps.ListKeyspacesRequest{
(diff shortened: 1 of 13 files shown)


pkg/engine/planetscale/progress.go
```diff
@@ -204,7 +228,8 @@ func (e *Engine) discoverMigrationContext(ctx context.Context, client psclient.P
 			continue
 		}
 		for _, r := range rows {
-			if r.MigrationContext != "" && !existingContexts[r.MigrationContext] {
+			_, exists := existingContexts[r.MigrationContext]
+			if r.MigrationContext != "" && !exists {
 				e.logger.Info("discovered schema change context", "context", r.MigrationContext)
 				return r.MigrationContext
```
Kiran01bm: seems like there will be ambiguity when more than one new context exists..
```
t0   Apply X (PR #50, table `orders`) on keyspace `commerce`
       captureExistingContexts → B = {ctx_old}
t1   X creates deploy request #50; Vitess is slow to expose its context
       discoverMigrationContext returns "" on all 10 attempts (the bug this PR fixes)
       → stored migration_context = "", baseline B persisted
t2   Apply Y (PR #51, table `inventory`) on the SAME keyspace `commerce`
       deploy request #51; its context ctx_Y becomes visible first
t3   X's Progress() recovery runs:
       SHOW VITESS_MIGRATIONS now shows ctx_Y (new) AND ctx_X (new)
       → returns ctx_Y  ❌  (first non-baseline row it hits)
t4   PR #50's progress UI now renders ctx_Y's data:
       table name `inventory`, ctx_Y's per-shard row counts, % done, ETA
```

pkg/engine/planetscale/progress.go
```diff
@@ -80,6 +80,15 @@ func (e *Engine) Progress(ctx context.Context, req *engine.ProgressRequest) (*en
 			}
 		}
 	}
+	if req.ResumeState.MigrationContext == "" && len(meta.ExistingMigrationCtxs) > 0 {
+		migrationContext := e.discoverMigrationContext(ctx, client, req.Database, req.Credentials, meta.ExistingMigrationCtxs)
+		if migrationContext != "" {
+			req.ResumeState = &engine.ResumeState{
+				MigrationContext: migrationContext,
+				Metadata:         req.ResumeState.Metadata,
+			}
```
Kiran01bm, line 89: perhaps we better persist the recovered number somewhere - ex: in a column that exists (migration_context) so that it turns the per-poll 9-query sweep into a one-time cost ?

example impact summary from my agent session:

```
Vitess DB with 8 keyspaces; Apply Z hit the late-context bug → stored migration_context = ""

Every CLI `schemabot status` / UI poll for Apply Z:
   BuildResumeState(vad)  →  migration_context still ""  (never persisted)
   → discoverMigrationContext:
        ListKeyspaces            (1 PlanetScale API call)
        SHOW VITESS_MIGRATIONS   × 8 keyspaces  (8 vtgate queries)
   → result thrown away, not saved
   → next poll repeats the same 9 queries
```

pkg/engine/planetscale/progress.go
aparajon: I _think_ this case should be fine, since PlanetScale only permits one deploy request to be in progress at a time for a database. So in that scenario above, `deploy request #51` wouldn't be able to start until `deploy request #50` reaches a terminal state

pkg/engine/planetscale/progress.go
aparajon: Good call — this is resolved now in `7db975b2`. `LocalClient.Progress` persists a recovered `ResumeState.MigrationContext` back into `vitess_apply_data.migration_context` when the stored value is missing/different, so the expensive context discovery sweep becomes a one-time recovery path instead of repeating on every poll. I also added a regression test for that path.

aparajon:
I checked compatibility with #242 by simulating a local merge of this branch into the opaque-resume-state branch. The two PRs are conceptually compatible, but this PR is not merge-compatible as-is.

The merge currently conflicts in:

- `pkg/storage/types.go`
- `pkg/tern/local_apply_grouped.go`
- `pkg/tern/local_client.go`
- `pkg/tern/local_client_test.go`
- `pkg/tern/local_control.go`

The main semantic overlap is the persistence path:

- This PR recovers a late Vitess `migration_context` and persists that recovery through `vitess_apply_data`.
- #242 makes `engine_resume_states` the durable source of truth for opaque engine state, and stops rebuilding engine state from `vitess_apply_data` for control/progress calls.

I think the engine-side idea in this PR is still the right follow-up, but the tern/storage side should be adapted after #242:

1. Keep the PlanetScale engine changes:
   - include the pre-deploy `existing_migration_contexts` baseline in `ResumeState.Metadata`
   - use deploy-request creation time to select the right newly observed Vitess context
   - return an updated `ResumeState` from `Progress` when the context is recovered

2. Drop or rewrite the `vitess_apply_data` durability changes:
   - no need for `existing_migration_contexts` in `vitess_apply_data` as an engine input
   - no need for `persistRecoveredVitessMigrationContext(...)`
   - #242 already persists updated `result.ResumeState` from progress polling into `engine_resume_states`

3. Keep `vitess_apply_data` projection-only if we still want operator/UI metadata there, but it should stay non-authoritative for engine calls.

Recommended order: merge #242 first, then rebase/rework this PR on top of it. That gives this PR a cleaner golden path:

```diagram
PlanetScale progress recovers migration_context
        │
        ▼
returns updated opaque ResumeState
        │
        ▼
SchemaBot persists it in engine_resume_states
        │
        ▼
future progress/control calls load engine_resume_states directly
```

PR #252, opened 2026-06-09
fix(github): respect repository schema path allowlists

aparajon:
## Summary

When a repository has server-configured `allowed_dirs`, treat those directories as the ownership boundary for SchemaBot config discovery. Configs discovered outside the managed paths are ignored instead of failing on an unrelated database name, while configs inside managed paths still fail closed when they cannot be routed safely.

This also adds operator-facing logs and status-check operation metrics for source-policy decisions during schema config discovery.

Generated with Amp.

### DIFF
--- a/pkg/webhook/schema_source_policy.go
+++ b/pkg/webhook/schema_source_policy.go
@@ -1,0 +1,85 @@
+package webhook
+
+import (
+	"context"
+
+	ghclient "github.com/block/schemabot/pkg/github"
+	"github.com/block/schemabot/pkg/metrics"
+)
+
+func (h *Handler) createManagedSchemaRequestFromPR(ctx context.Context, client *ghclient.InstallationClient, repo string, pr int, environment, databaseName, source string) (*ghclient.SchemaRequestResult, error) {
+	schemaResult, err := client.CreateSchemaRequestFromPR(ctx, repo, pr, environment, databaseName)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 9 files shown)


pkg/webhook/schema_source_policy.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/block/schemabot/pkg/metrics"
+)
+
+func (h *Handler) createManagedSchemaRequestFromPR(ctx context.Context, client *ghclient.InstallationClient, repo string, pr int, environment, databaseName, source string) (*ghclient.SchemaRequestResult, error) {
+	schemaResult, err := client.CreateSchemaRequestFromPR(ctx, repo, pr, environment, databaseName)
+	if err != nil {
+		return nil, err
+	}
+	if !h.schemaPathManagedByRepo(ctx, repo, pr, schemaResult.HeadSHA, schemaResult.Database, schemaResult.Type, schemaResult.SchemaPath, source) {
+		return nil, ghclient.ErrNoConfig
```
Kiran01bm, line 16: **nit** - `schemaPathManagedByRepo` returning true in error cases (config unavailable, db not configured) means the name is reads as "is managed" but actually means "should proceed", but `shouldProcessSchemaConfig` or similar would describe the contract better.

Kiran01bm (approved):
left a minor naming nit.. but rest lgtm..

PR #253, opened 2026-06-10
feat(operator): derive applies.state from child apply_operations

Kiran01bm:
## What and Why ?

Invert the apply-state source of truth in the flag-gated operator path. After the operator drives an `apply_operations` row to its state, `applies.state` is now **derived** from the apply's child operation rows via the same `state.DeriveApplyState` already used to derive apply state from task states — rather than only being mirrored down onto the operation row.

```
ResumeApply writes durable applies.state
│
▼
operator reloads parent apply
│
▼
markOperationFromApplyState → child operation row reaches its state
│  (reports updated=true only on a durable terminal write)
▼
updateApplyStateFromOperations → re-derive applies.state from ALL
child operations, persist only if it differs
```
Gated behind `operator_claim_operations` (default off); the legacy `FindNextApply` path and `ResumeApply` are untouched. While an apply owns exactly one operation the derived value equals what `ResumeApply` already wrote, so this is a no-op in value — it establishes the aggregation seam the multi-deployment fan-out needs.

Fail-closed: skips derivation when the child write failed (never aggregates a stale child state), refuses to revive a terminal apply from inconsistent children, and errors on an apply with zero operation rows. `markOperationFromApplyState` now returns `(updated bool, err error)` so callers sequence the derivation correctly.

### DIFF
--- a/integration/operator_test.go
+++ b/integration/operator_test.go
@@ -264,6 +264,15 @@ func TestOperator_OperatorClaimsOperationToCompletion(t *testing.T) {
 	assert.Equal(t, state.ApplyOperation.Completed, ops[0].State, "operator marks the claimed operation completed")
 	assert.Equal(t, apply.Deployment, ops[0].Deployment)
 	require.NotNil(t, ops[0].CompletedAt, "completed operation stamps completed_at")
+
+	// applies.state is derived from its child operation rows: the parent must
+	// equal DeriveApplyState of the operation states it owns.
+	finalApply, err := ts.Storage.Applies().Get(ctx, apply.ID)
+	require.NoError(t, err)
+	require.NotNil(t, finalApply)
+	derived := state.DeriveApplyState([]string{ops[0].State})
+	assert.Equal(t, state.Apply.Completed, derived)
+	assert.Equal(t, derived, finalApply.State, "parent apply state is derived from its child operations")
 }
 
 // TestOperator_OperatorReconcilesOperationWhenParentTerminal covers the safety
@@ -314,6 +323,15 @@ func TestOperator_OperatorReconcilesOperationWhenParentTerminal(t *testing.T) {
 	require.NotNil(t, op)
 	assert.Equal(t, state.ApplyOperation.Completed, op.State)
 	require.NotNil(t, op.CompletedAt, "reconciled completed operation stamps completed_at")
+
+	// Even on the terminal-parent reconciliation path, the parent stays equal to
+	// DeriveApplyState of its child operations.
+	parent, err := ts.Storage.Applies().Get(ctx, applyID)
+	require.NoError(t, err)
+	require.NotNil(t, parent)
+	derived := state.DeriveApplyState([]string{op.State})
+	assert.Equal(t, state.Apply.Completed, derived)
+	assert.Equal(t, derived, parent.State, "parent apply state is derived from its child operations")
 }
 
 func TestOperator_ClaimOrdering(t *testing.T) {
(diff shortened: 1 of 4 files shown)


aparajon (approved):

aparajon:
One possible edge case to double-check before enabling the operation-claim path broadly: stopped applies with a pending start request.

At the apply level, `FindNextApply` treats `state = stopped` plus a pending `start` control request as claimable, so the operator can resume the apply and process the start request. In the operation-claim path, `markOperationFromApplyState` mirrors a stopped parent into the child `apply_operations` row, but `FindNextApplyOperation` still appears to skip stopped operation rows because they are terminal from the operation-claim query's perspective.

That seems like it could strand this sequence under `operator_claim_operations`:

1. operation is claimed and driven
2. user stops the apply
3. parent apply becomes `stopped`
4. child operation is mirrored to `stopped`
5. user sends start
6. parent has a pending start request, but the stopped child operation is not claimable

Maybe this is intentionally deferred until the native operation-scoped resume path replaces the bridge logic. If so, it may be worth either keeping this flag away from stop/start flows until then, or adding parity in `FindNextApplyOperation` for stopped child operations whose parent apply has a pending start request.

Generated by Amp.

Kiran01bm:
> One possible edge case to double-check before enabling the operation-claim path broadly: stopped applies with a pending start request.
> 
> At the apply level, `FindNextApply` treats `state = stopped` plus a pending `start` control request as claimable, so the operator can resume the apply and process the start request. In the operation-claim path, `markOperationFromApplyState` mirrors a stopped parent into the child `apply_operations` row, but `FindNextApplyOperation` still appears to skip stopped operation rows because they are terminal from the operation-claim query's perspective.
> 
> That seems like it could strand this sequence under `operator_claim_operations`:
> 
> 1. operation is claimed and driven
> 2. user stops the apply
> 3. parent apply becomes `stopped`
> 4. child operation is mirrored to `stopped`
> 5. user sends start
> 6. parent has a pending start request, but the stopped child operation is not claimable
> 
> Maybe this is intentionally deferred until the native operation-scoped resume path replaces the bridge logic. If so, it may be worth either keeping this flag away from stop/start flows until then, or adding parity in `FindNextApplyOperation` for stopped child operations whose parent apply has a pending start request.
> 
> Generated by Amp.

good catch — this is a real gap - under `operator_claim_operations` the loop runs `recoverApplyOperation` and returns early, so the apply-level `FindNextApply` (which does treat stopped + pending start as claimable) is bypassed entirely as `FindNextApplyOperation` had no equivalent clause, so the stopped child stranded the start request permanently (the stale-active branch can't rescue it — stopped isn't an active state).

went with option (b): `FindNextApplyOperation` now mirrors `FindNextApply`'s stop/start rule — a stopped operation whose parent apply has a pending start request is reclaimable, with no deployment-order gate (resuming a stopped row is recovering work it already started, like the stale-active clause) and the re-leased row keeps its stopped state and the resume drive transitions it - added integration coverage for the claim plus the pending-vs-completed request filter - flag stays default-off.

also ive done a thorough claim dis-agreement table below and found a couple of more gaps (documented in the table below)
**note** - the `failed_retryable` and `pending` + `start-request` rows are still divergent — out of scope here but will check and do a follow-up

| Claimable? | `FindNextApply` (apply-level) | `FindNextApplyOperation` (operation-level, before fix) |
|---|---|---|
| `pending` + tasks | ✅ | ✅ (`state = pending`) |
| stale active | ✅ | ✅ (`state IN activeStates AND stale`) |
| `failed_retryable` | ✅ | ❌ |
| `pending` + pending start req | ✅ | ❌ |
| **`stopped` + pending start req** | ✅ | ❌ → ✅ after this fix |

PR #256, opened 2026-06-10
feat(storage): add Tasks().GetByApplyOperationID

Kiran01bm:
## What and Why ?

Add a storage read primitive, `Tasks().GetByApplyOperationID(ctx, operationID)`, that returns the tasks belonging to a single `apply_operation` — one deployment of a multi-deployment apply — using the `idx_apply_operation_id` index.

Today the operator drives an apply by loading **all** of its tasks (`GetByApplyID`). To eventually drive one deployment independently of its siblings, a worker needs to load just that operation's tasks. This PR adds that primitive with **no callers** — the smallest, dormant first step.

It is behaviour-identical today: while the config layer hard-blocks more than one deployment per environment, an apply's single operation owns all of its tasks, so this returns the same rows as `GetByApplyID`. It only diverges once an apply fans out across deployments.

The method deliberately does **not** fall back to the parent apply's tasks on an empty result — an empty slice is the correct fail-closed signal for future operation-scoped callers.

## Changes
- `TaskStore.GetByApplyOperationID` interface method + `mysqlstore` implementation.
- Test: `TestTaskStore_GetByApplyOperationID` (per-operation isolation; empty operation returns empty, no fallback).

## Where this sits — operation-scoped engine drive

The operator should drive only the deployment a worker has claimed, loading tasks by `apply_operation_id` instead of resuming the whole apply. Because the resume path is shared by the local and remote clients and re-queries tasks in many places, that lands as a sequence of dormant PRs — only the final switch changes behaviour:

| # | PR | Purpose | Builds on |
|---|---|---|---|
| 1 | **this PR** — `Tasks().GetByApplyOperationID` | Read primitive for one operation's tasks; no callers; no fallback on empty. | — |
| 2 | LocalClient operation-scoped resume | `LocalClient.ResumeApplyOperation` loads tasks via the primitive; legacy `ResumeApply` untouched. | 1 |
| 3 | GRPCClient operation-scoped resume | `GRPCClient.ResumeApplyOperation`; thread the scope through the remote drive's re-query sites. | 1, mirrors 2 |
| 4 | Promote to the `Client` interface | Add `ResumeApplyOperation` once both clients implement it. | 2 + 3 |
| 5 | Switch the operation claim loop | Operator drives the claimed operation, fail-closed on zero tasks; legacy apply path unchanged. | 4 |

Every step is dormant under the single-deployment hard-block — an operation-scoped load returns the same rows as the apply-scoped load until an apply fans out across deployments.

### DIFF
--- a/pkg/storage/mysqlstore/tasks.go
+++ b/pkg/storage/mysqlstore/tasks.go
@@ -164,9 +164,9 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
 func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID int64) ([]*storage.Task, error) {
 	rows, err := s.db.QueryContext(ctx, `
 		SELECT `+taskColumns+`
-		FROM `+"`tasks`"+` FORCE INDEX (`+"`idx_apply_operation_id`"+`)
-		WHERE `+"`apply_operation_id`"+` = ?
-		ORDER BY `+"`created_at`"+` DESC
+		FROM tasks FORCE INDEX (idx_apply_operation_id)
+		WHERE apply_operation_id = ?
+		ORDER BY created_at DESC, id DESC
 	`, applyOperationID)
 	if err != nil {
 		return nil, fmt.Errorf("query tasks for apply_operation %d: %w", applyOperationID, err)
--- a/pkg/storage/mysqlstore/tasks.go
+++ b/pkg/storage/mysqlstore/tasks.go
@@ -156,6 +156,26 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
 	return scanTasks(rows)
 }
 
+// GetByApplyOperationID returns the tasks for a single apply_operation (one
+// deployment of a multi-deployment apply). While the config layer hard-blocks
+// more than one deployment per environment this returns the same rows as
+// GetByApplyID for the apply's single operation; it lets a worker drive and
+// reconcile one deployment independently once an apply fans out across several.
+func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID int64) ([]*storage.Task, error) {
+	rows, err := s.db.QueryContext(ctx, `
+		SELECT `+taskColumns+`
+		FROM `+"`tasks`"+` FORCE INDEX (`+"`idx_apply_operation_id`"+`)
+		WHERE `+"`apply_operation_id`"+` = ?
+		ORDER BY `+"`created_at`"+` DESC
+	`, applyOperationID)
+	if err != nil {
+		return nil, fmt.Errorf("query tasks for apply_operation %d: %w", applyOperationID, err)
+	}
+	defer utils.CloseAndLog(rows)
+
+	return scanTasks(rows)
+}
+
 // GetByDatabase returns all tasks for a database.
 // Results are ordered by created_at DESC, then by id DESC as a tiebreaker
 // (since created_at only has second precision).
--- a/pkg/storage/mysqlstore/tasks.go
+++ b/pkg/storage/mysqlstore/tasks.go
@@ -164,7 +164,7 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
 func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID int64) ([]*storage.Task, error) {
 	rows, err := s.db.QueryContext(ctx, `
 		SELECT `+taskColumns+`
-		FROM tasks FORCE INDEX (idx_apply_operation_id)
+		FROM tasks
 		WHERE apply_operation_id = ?
 		ORDER BY created_at DESC, id DESC
 	`, applyOperationID)
--- a/pkg/storage/mysqlstore/tasks.go
+++ b/pkg/storage/mysqlstore/tasks.go
@@ -173,7 +173,16 @@ func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID 
 	}
 	defer utils.CloseAndLog(rows)
 
-	return scanTasks(rows)
+	tasks, err := scanTasks(rows)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/storage/mysqlstore/tasks.go
```diff
@@ -156,6 +156,26 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
...
+// deployment of a multi-deployment apply). While the config layer hard-blocks
+// more than one deployment per environment this returns the same rows as
+// GetByApplyID for the apply's single operation; it lets a worker drive and
+// reconcile one deployment independently once an apply fans out across several.
+func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID int64) ([]*storage.Task, error) {
+	rows, err := s.db.QueryContext(ctx, `
+		SELECT `+taskColumns+`
+		FROM `+"`tasks`"+` FORCE INDEX (`+"`idx_apply_operation_id`"+`)
+		WHERE `+"`apply_operation_id`"+` = ?
+		ORDER BY `+"`created_at`"+` DESC
```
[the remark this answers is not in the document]
Kiran01bm: good point - i think we need to add this as a agent instruction somewhere as its not the 1st time we've had this gap surfacing!

this is fixed in https://github.com/block/schemabot/pull/256/changes/7f4f5a84416a54afba4888a51237a234a9003dd9

pkg/storage/mysqlstore/tasks.go
```diff
@@ -156,6 +156,26 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
...
+	`, applyOperationID)
+	if err != nil {
+		return nil, fmt.Errorf("query tasks for apply_operation %d: %w", applyOperationID, err)
+	}
+	defer utils.CloseAndLog(rows)
+
+	return scanTasks(rows)
```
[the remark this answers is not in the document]
Kiran01bm: for sure  - its now addressed in https://github.com/block/schemabot/pull/256/changes/f4f950e0dc815b88edf9618dd7e2e3013bf653bc

pkg/storage/mysqlstore/tasks_test.go
```diff
@@ -0,0 +1,86 @@
...
+	require.Len(t, tasksB, 1)
+	assert.Equal(t, "task_b_users", tasksB[0].TaskIdentifier)
+	require.NotNil(t, tasksB[0].ApplyOperationID)
+	assert.Equal(t, opB, *tasksB[0].ApplyOperationID)
+
+	// An operation with no tasks returns an empty slice — no fallback to the
+	// parent apply's tasks.
+	tasksNone, err := store.Tasks().GetByApplyOperationID(ctx, opB+1000)
+	require.NoError(t, err)
+	assert.Empty(t, tasksNone)
```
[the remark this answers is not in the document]
Kiran01bm: the tests are now modified to have the real op record and query for it - thank you - https://github.com/block/schemabot/pull/256/changes/f4f950e0dc815b88edf9618dd7e2e3013bf653bc

pkg/storage/mysqlstore/tasks.go
```diff
@@ -156,6 +156,35 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
...
 
+// GetByApplyOperationID returns the tasks for a single apply_operation (one
+// deployment of a multi-deployment apply). While the config layer hard-blocks
+// more than one deployment per environment this returns the same rows as
+// GetByApplyID for the apply's single operation; it lets a worker drive and
+// reconcile one deployment independently once an apply fans out across several.
+func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID int64) ([]*storage.Task, error) {
+	rows, err := s.db.QueryContext(ctx, `
+		SELECT `+taskColumns+`
+		FROM tasks FORCE INDEX (idx_apply_operation_id)
```
aparajon: Do we need to do a `force index` here?

pkg/storage/mysqlstore/tasks.go
```diff
@@ -156,6 +156,35 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
...
+	defer utils.CloseAndLog(rows)
+
+	tasks, err := scanTasks(rows)
+	if err != nil {
+		return nil, err
+	}
+	if tasks == nil {
+		// Return a non-nil empty slice so callers can never confuse "operation
+		// has no tasks" with nil and fall back to the parent apply's tasks.
+		return []*storage.Task{}, nil
```
aparajon, line 183: Thoughts on making this an error?

pkg/storage/mysqlstore/tasks.go
Kiran01bm: good pick-up - i dont think so and i'll remove it as the optimizer should reliably pick the index for a single-column equality lookup, the hint was inconsistent with GetByApplyID, moreover it would turn an index rename/drop into a query error instead of a fallback scan.

pkg/storage/mysqlstore/tasks.go
Kiran01bm: actually i reasoned about it in response to the copilots suggestion - https://github.com/block/schemabot/pull/256#discussion_r3386942196 and the reasoning was

A storage read primitive should treat "no rows" as data, not failure. Making it an error would:

- Conflate causes. The caller can no longer distinguish "query failed" from "operation legitimately has zero tasks" without string-matching the error — directly against the AGENTS.md rule "Do not conflate causes in one branch."
- Put policy in the wrong layer. Whether zero-tasks-is-fatal is a drive-flow decision, not a storage one. And it's already enforced at the right layer with richer context — dispatchPendingApply does if len(tasks) == 0 { markRemoteApplyFailed(..., "no tasks found") }, which logs the apply identifier, sets the retryable flag, and fails closed. A bare error from storage couldn't carry any of that.
- Break benign callers. Counting/iterating/progress callers that find an empty operation would have to special-case an error for a normal state.

aparajon (approved):

PR #264, opened 2026-06-10
fix(github): enforce actor authorization for rollback and unlock commands

morgo:
When PR command actor authorization is enabled, the `rollback`, `rollback-confirm`, and `unlock` PR comment commands skipped the allowlist check that gates `apply`, `apply-confirm`, `stop`, and `cutover`. Any GitHub user who could comment on a PR could execute rollback DDL (which forces `allow_unsafe`) or force-release schema change locks, including CLI-owned locks.

- `rollback` and `rollback-confirm` now enforce the fail-closed actor authorization gate before any lock is acquired or released and before any plan or apply executes.
- `unlock` enforces the gate for every matched lock's database before releasing anything; locks are not environment-scoped, so authorization is enforced per database.
- The conditional GitHub-client resolution used by the gate is extracted into a shared helper (`actorAuthorizationClient`) and reused by `stop` and `cutover`.

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -4302,6 +4302,21 @@ A configured SchemaBot admin/database operator should inspect SchemaBot authoriz
 </details>
 
 <details>
+<summary><a name="actor-authorization-database-not-configured"></a><strong>Actor Authorization: Database Not Configured</strong></summary>
+
+
+## SchemaBot Command Not Authorized
+
+**Database**: `payments`
+
+`schemabot unlock` cannot run because database `payments` is not configured on this SchemaBot instance.
+
+Verify the database name, or run the command against the SchemaBot instance that manages this database.
+
+
+</details>
+
+<details>
 <summary><a name="cutover-command-accepted"></a><strong>Cutover Command Accepted</strong></summary>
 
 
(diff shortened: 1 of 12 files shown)


aparajon:
Verified the safety-critical properties on the head: authorization runs before any side effect on all seven mutating commands (rollback before lock reads and plan execution; rollback-confirm independently gated, so allowlist revocation between rollback and confirm still blocks; unlock gated per database, all-or-nothing). Fail-closed holds across client-resolution failure, nil client with team policies, empty allowlist, and missing actor. Actor identity comes from the HMAC-verified webhook payload, not comment content. Tests assert denial side effects (no lock acquired/released), not just comments.

Three nits:

1. **Audit logging on the allow path** — denials log at Warn with full identifiers, but *allowed* rollback/unlock log only at Debug. For commands that execute DDL or force-release locks, the allow decision is audit-relevant; suggest Info with the same identifiers.

2. **Information probe before the gate** — `handleRollbackCommand` posts "apply not found" / lock-conflict details before authorization, so unauthorized commenters can probe apply IDs. Low risk, but consider gating earlier once the database is known.

3. **Comment differentiation** — unlock of an unconfigured database renders the same comment as a plain authorization denial; logs/metrics differentiate, but a hint ("database is not configured on this instance") would save operators a round-trip.

Verdict: LGTM with those nits.

---

aparajon (approved):
Approving — authorization verified to run before any side effect on all seven mutating commands, fail-closed across client/allowlist/actor failure modes, actor identity from the HMAC-verified payload. Audit-logging nits are non-blocking.

PR #268, opened 2026-06-10
fix(operator): write apply_operations row when creating applies via the Tern client

morgo:
`LocalClient.Apply` (the path used when SchemaBot serves as a remote Tern engine over gRPC) created the `applies` row and its tasks separately and never wrote an `apply_operations` child row, so every task carried a nil `ApplyOperationID`. The HTTP/API apply path already dual-writes that row via `CreateWithTasksAndOperations`; the Tern-client path did not. Two consequences:

- With operator operation-claiming enabled, the claim loop selects work exclusively from `apply_operations`. Applies created without a child row would never be started, crash-recovered, or retried — silently.
- Grouped applies require a non-nil `ApplyOperationID` to persist engine resume state. The apply was failed *after* the deploy request had already been created, orphaning the remote deploy request.

`Apply` now uses `CreateWithTasksAndOperations` to write the applies row, its tasks, and exactly one `apply_operations` row in a single transaction, mirroring the API path. The store links each task to the operation via `ApplyOperationID`. The sequential (Spirit) path is unaffected by the linkage and continues to run unchanged.

Legacy applies created before this dual-write existed still lack an operation row; backfilling them is a possible follow-up and is intentionally out of scope here.

### DIFF
--- a/pkg/tern/local_client.go
+++ b/pkg/tern/local_client.go
@@ -611,6 +611,11 @@ func (c *LocalClient) Apply(ctx context.Context, req *ternv1.ApplyRequest) (*ter
 	// claimable, resumable operation. CreateWithTasksAndOperations links each
 	// task to the single operation via ApplyOperationID, which the engine
 	// resume-state path requires and the operator claim loop selects on.
+	//
+	// CutoverPolicy and HaltOnFailure are intentionally left unset: the Tern
+	// client has no environment config to resolve them from (unlike the API
+	// apply path), so the store applies its safe defaults (rolling cutover,
+	// halt on failure).
 	operations := []*storage.ApplyOperation{{
 		Deployment: apply.Deployment,
 		Target:     plan.Target,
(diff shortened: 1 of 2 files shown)


aparajon:
Verified `CreateWithTasksAndOperations` inserts the apply, the single operation, and tasks in one transaction with the same active-apply gate the old path ran — a strict improvement over the previous non-atomic apply-row + per-task create loop, which could leave a partially populated apply on crash. The 1:1 apply:operation invariant matches the current model, and the store's multi-op guard will hard-fail if that assumption is ever lifted without caller changes.

Two small items:

1. The operation row intentionally omits `CutoverPolicy`/`HaltOnFailure` (the store defaults them), while the API path resolves both from server config. That asymmetry is reasonable — this client has no environment config to resolve from — but a one-line comment on the `operations` literal stating the defaults are intentional would keep it from reading as an omission.

2. The base is a bit behind main (recent changes touched this file); a quick rebase before merge so CI runs against the operation-scoped resume code would be good. No textual conflicts expected.

Verdict: LGTM with those nits.

---

aparajon (approved):

aparajon (approved):
Approving — atomic apply+operation+tasks creation is a strict improvement, 1:1 invariant holds with a hard-fail guard. Please rebase before merge so CI runs against the operation-scoped resume code.

PR #271, opened 2026-06-10
fix(operator): stop conflict check from failing stopped or retryable tasks

morgo:
`tryResolveStaleTask` runs on every apply (via the active-task conflict check) for each non-terminal task on the target database, and converted any task whose engine reported `No active schema change` to `failed`. But Spirit reports exactly that whenever it has no running schema change — the normal resting state for a deliberately **stopped** task (checkpoint preserved) and a **failed_retryable** task awaiting operator retry. Because the check sees every task on the database, any new apply silently failed those resting tasks with a misleading "server may have crashed" message, voiding resumes and the operator retry budget.

Now only genuinely in-flight tasks (running, waiting_for_cutover, cutting_over, waiting_for_deploy, recovering) whose engine has gone silent are treated as abandoned. Resting `stopped`/`failed_retryable` tasks are left intact, and the existing conflict logic decides whether the new apply may proceed rather than running over paused work.

### DIFF
--- a/pkg/state/task.go
+++ b/pkg/state/task.go
@@ -61,6 +61,21 @@ func IsTerminalTaskState(s string) bool {
 	}
 }
 
+// IsInFlightTaskState returns true if the state implies the engine is actively
+// working on the task. These are the only states for which a missing engine
+// migration means the task was abandoned (e.g. a server crash). Resting states
+// such as Stopped and FailedRetryable also have no active engine work, but that
+// absence is expected — Spirit keeps the checkpoint while the operator decides
+// whether to resume or retry — so they are excluded here.
+func IsInFlightTaskState(s string) bool {
+	switch NormalizeState(s) {
+	case Task.Running, Task.WaitingForCutover, Task.CuttingOver, Task.WaitingForDeploy, Task.Recovering:
+		return true
+	default:
+		return false
+	}
+}
+
 // NormalizeTaskStatus maps a raw engine status to a canonical Task state.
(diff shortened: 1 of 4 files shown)


aparajon:
Verified the `IsInFlightTaskState` set matches the stop path's live-work set exactly, and preserving resting tasks makes the conflict check *stricter*, not looser — the stopped/retryable task now blocks the new apply rather than being destroyed to admit it. That's the safer direction.

On the open question of adding `RevertWindow` to the in-flight set: recommend **against**. `tryResolveStaleTask` probes the engine with no resume state, and the PlanetScale engine's `Progress` reports "no active schema change" whenever resume metadata is absent — so an in-flight-classified revert-window task would get marked `failed`, recording a *deployed* change as failed. Revert-window tasks are deliberately operator-owned; the test pinning `RevertWindow` in the resting list is correct as written.

One follow-up worth its own PR: "engine has gone silent" in `tryResolveStaleTask` is judged against the local pod's in-memory engine. A running task driven by a healthy peer (lease held, heartbeat fresh) looks abandoned from another pod and gets failed — a stale-worker-overwrites-newer-state window. This PR shrinks the blast radius, but the in-flight branch should also require a stale apply heartbeat before declaring abandonment. The `result.State.IsTerminal()` branch above the new guard has a narrower version of the same exposure.

Verdict: LGTM; follow-ups noted.

---

aparajon (approved):
Approving — preserving stopped/retryable tasks makes the conflict check stricter, not looser. Per my earlier comment, recommend keeping RevertWindow in the resting list as written.

PR #282, opened 2026-06-10
fix(ci): make E2E image builds resilient to registry flakiness

morgo:
## Summary

E2E jobs (E2E MySQL, E2E Vitess, E2E K8s) build images that pull `alpine:3.19` and `mysql:8.0` from Docker Hub at build/deploy time. These are **unauthenticated** pulls with **no retry**, which produces the recurring intermittent failures:

- `Get "https://registry-1.docker.io/v2/": net/http: request canceled (Client.Timeout exceeded while awaiting headers)` and `failed to resolve source metadata for docker.io/library/alpine:3.19` — Docker Hub rate-limit / timeout on the live pull.
- `failed to execute bake: read |0: file already closed` — the buildx progress race that surfaces when a base-image pull stalls in the middle of `docker compose up --build`.

The base image references themselves are fine; the problem is that the pull happens live, on the build/deploy critical path, with no resilience.

## Fix

Pre-pull the Docker Hub base images with bounded retry and exponential backoff **before** the build/deploy step, so the build resolves them from the local image cache instead of issuing live registry pulls mid-build.

- New reusable composite action `.github/actions/prepull-base-images` (used by the MySQL and Vitess jobs, which build via `docker compose up --build` against the host daemon).
- The K8s job pre-pulls inline inside the existing `minikube docker-env` context so it warms minikube's daemon — both the `docker build` of `Dockerfile.dev` and kubelet's `mysql:8.0` pod pull (`imagePullPolicy: IfNotPresent`) then hit the local cache.

This does not mask test failures or touch any test timeouts: a genuine, persistent registry outage still fails the job loudly after exhausting retries; only transient timeouts are absorbed.

```
checkout ──> pre-pull base images (retry/backoff) ──> build/up --build ──> test
                       │
                       └─ warms local docker cache so the build/deploy
                          step resolves alpine:3.19 / mysql:8.0 locally
                          instead of a live Docker Hub pull
```

### DIFF
--- a/.github/actions/prepull-base-images/action.yml
+++ b/.github/actions/prepull-base-images/action.yml
@@ -1,0 +1,54 @@
+name: "Pre-pull base images"
+description: >-
+  Pre-pull the Docker Hub base images that E2E image builds depend on, with
+  bounded retry and backoff. Warming the local image cache before
+  `docker compose up --build` / `docker build` means the subsequent build
+  resolves base images locally instead of issuing live Docker Hub pulls
+  mid-build, where unauthenticated rate-limit timeouts and the buildx progress
+  race ("read |0: file already closed") surface.
+
+inputs:
+  images:
+    description: "Space- or newline-separated list of base images to pull."
+    required: false
+    default: |
+      alpine:3.19
+      mysql:8.0
+  attempts:
+    description: "Maximum pull attempts per image."
+    required: false
+    default: "5"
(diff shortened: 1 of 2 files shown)


morgo:
I think it's cleaner to not pull this. It's only been flaky lately.. lets give it a chance to recover. We can revisit if that proves untrue.

PR #300, opened 2026-06-11
fix(github): block applies when a required check has not reported

morgo:
When `required_checks` is configured, the passing-checks apply gate only evaluated the required checks GitHub had already reported on the PR commit. A configured required check that had not reported yet neither failed nor counted as in-progress, so the apply could proceed before that check ever ran — a fail-open gap in a tier-0 safety gate.

This treats every configured required check that is absent from the fetched statuses as a not-yet-reported in-progress blocker, mirroring how a legacy commit status in the `EXPECTED` state maps to `in_progress`. The gate now fails closed and the apply-blocked comment names the missing required check(s) so the operator knows what still has to report.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -623,6 +623,9 @@ func (c *ServerConfig) Validate() error {
 	if err := c.validateGitHubAppsConfig(); err != nil {
 		return err
 	}
+	if err := c.validateRequiredChecksNotAggregate(); err != nil {
+		return err
+	}
 	if c.PendingDropsEnabled() {
 		if _, err := c.PendingDropsRetention(); err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


aparajon:
The fail-closed mechanics are right: `missingRequiredChecks` only runs after a successful status fetch, the three causes (API/permission failure, failing checks, never-reported checks) keep separate branches with distinct handling, no path converts an API error into a pass, and the REST fetch paginates fully. Name matching is consistent with `IsCheckRequired`. The safety direction is correct — but the operator-facing message needs a change before merge:

1. **A never-reported check renders as "still running… wait and retry"** — for the realistic causes (typo in `required_checks`, check not configured on the repo, path-filtered workflow), the check will *never* report, and this advice sends an operator into an indefinite wait. The table's "not reported" row helps, but the remediation text is wrong for those rows. Suggest a conditional paragraph when missing checks exist: "`<name>` has not reported on this commit. If it never reports, verify the name in `required_checks` matches the check exactly and that it runs on this PR."

2. **Log the missing names** — "apply blocked by in-progress PR checks" with a `missing_required_count` conflates two causes in one message. When missing checks are the blocker, log them by name so triage doesn't require reproducing the gate.

3. Nit: a SchemaBot-owned check counts as "reported" here (correct — avoids self-deadlock), but an aggregate-check entry in `required_checks` is then silently unenforced by this gate. Consider config validation rejecting it explicitly.

Verdict: needs changes for the message; mechanics are sound.

---

aparajon (approved):

PR #309, opened 2026-06-11
feat(github): actionable check-trust failures and chain trust model

aparajon:
## Summary

SchemaBot gates trust check runs **only by the creating GitHub App, never by check name** — in both directions. The promotion gate accepts a prior environment's aggregate check only from a trusted chain App (fails closed otherwise), and the `require_passing_checks` gate excludes only trusted Apps' aggregates from CI-health blocking. A same-named check from any other app cannot satisfy the promotion gate and cannot be hidden from the CI gate.

`trusted-check-app-slugs` is documented as chain-level shared configuration: every deployment in a promotion chain carries the same complete list of the chain's App slugs (like `environment_order`). That removes the cross-deployment blocking hazard — a sibling's `action_required` aggregate is expected workflow state, not failing CI — without introducing any name-trusted path.

```text
check on PR head               promotion gate      passing-checks gate
-----------------------------  ------------------  -------------------------------
trusted chain App aggregate    can satisfy         excluded (SchemaBot state)
untrusted app, aggregate name  cannot satisfy;     blocks if not passing
                               blocked comment     + warn log + metric
                               names the app
any other external check       cannot satisfy      blocks if not passing
```

**Actionable failure UX.** The promotion gate now distinguishes "no prior-environment check exists" from "a check exists but its App is not trusted", because the remediations differ. The first comment directs the user to plan/apply the prior environment; the second names the untrusted app(s) and directs an operator to `trusted-check-app-slugs` (or to investigate an impersonating check) instead of suggesting a re-plan that cannot change check ownership. `FindCheckRunByName` reports the ignored apps to make this possible.

**Observability.** New counter `schemabot.untrusted_aggregate_named_checks_total` (repository, environment, app_slug, gate) plus warn logs with the creating app slug, in both the passing-checks and promotion gates — so operators can tell a check-name spoof attempt from a chain member missing from the trusted list. `PRCheckStatus` now carries the creating App slug.

**Docs.** `check-runs.md` adds a cross-deployment trust model: prior-environment checks are **sequencing evidence, never execution authority** — a tampered earlier environment can defeat staging-first ordering but cannot influence DDL content, actor authorization, review gating, or lint, all evaluated locally by the later environment. Branch-protection setup now recommends pinning required SchemaBot checks to their publishing App, closing the one name-trust surface SchemaBot cannot enforce itself.

---

*This PR was generated by [Amp](https://ampcode.com) with Fable (claude-fable-5).*

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -8,6 +8,7 @@
   - [Building DSNs from separate secrets](#building-dsns-from-separate-secrets)
 - [gRPC Mode](#grpc-mode)
 - [Multi-Deployment Environment (preview)](#multi-deployment-environment-preview)
+  - [Deployment Order](#deployment-order)
 - [Environment Order](#environment-order)
 - [Hybrid Mode](#hybrid-mode)
 - [Operator Workers](#operator-workers)
@@ -469,6 +470,9 @@ github:
   private-key: "file:/run/secrets/staging-github-key.pem"
   webhook-secret: "env:STAGING_WEBHOOK_SECRET"
   check-name: "SchemaBot X"
+  trusted-check-app-slugs:  # the chain's App slugs — identical in every instance
+    - schemabot-staging
+    - schemabot-production
 ```
 
 ### Production Instance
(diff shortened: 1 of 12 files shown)


pkg/webhook/apply_gating.go
```diff
@@ -136,8 +151,31 @@ func (h *Handler) githubAppDisplayNameForRepo(repo string, client *ghclient.Inst
 	return ""
 }
 
+// flagUntrustedAggregateNamedChecks surfaces PR checks that carry this
+// deployment's aggregate Check Run name but were not created by a trusted
+// SchemaBot GitHub App. Such a check is either a spoof attempt or a sibling
+// deployment whose App slug is missing from trusted-check-app-slugs; in both
+// cases it is treated as ordinary external CI (so a non-passing one blocks
+// applies), and operators need the app identity to tell the two apart.
+func (h *Handler) flagUntrustedAggregateNamedChecks(ctx context.Context, statuses []ghclient.PRCheckStatus, repo string, pr int, headSHA, environment string) {
```
Kiran01bm: `filterNonPassingNonSchemaBotChecks` skips non-required checks when in required-checks mode ([apply_gating.go#L38](file:///Users/kmuddukrishna/code/schemabot/pkg/webhook/apply_gating.go#L38)), so a non-required untrusted aggregate won't actually block but `flagUntrustedAggregateNamedChecks` has no such filter, so it still logs `...will block applies unless passing` and increments the metric for a check that won't block in that mode..

docs/configuration.md
Kiran01bm, line 1: something my agent surfaced - 

Stale config field doc comment (should-fix)

Explanation: The TrustedCheckAppSlugs doc comment only describes the promotion-gate purpose ("verify a prior environment's aggregate"). It omits the new passing-checks-gate purpose that this PR's chain-level model depends on: every instance must list the whole chain (e.g. staging must trust schemabot-production) so a sibling's action_required aggregate isn't treated as failing external CI. As written, an operator following the schema comment under-configures and hits the deadlock this PR fixes.

Inject at: [pkg/api/config.go#L236-L244](https://github.com/block/schemabot/blob/main/pkg/api/config.go#L236-L244) — the // TrustedCheckAppSlugs lists ... comment block. Add a sentence that it is chain-level shared config (same complete list in every instance, own slug included), used by both the promotion gate and the passing-checks gate. Point to docs/configuration.md.

Kiran01bm (approved):

pkg/webhook/apply_gating.go
aparajon: Addressed in 6dc7d8c5 — the helper now computes `required_checks` narrowing and logs the actual gating impact: a non-required untrusted aggregate-named check is reported as "required_checks narrowing keeps it from gating applies, but it may be impersonating SchemaBot", while gating checks keep the "will block applies unless passing" message. The counter intentionally still fires in both cases — it is an identity-mismatch/spoof signal rather than a blocking signal — and the metric doc comment now states that explicitly. Added `TestFlagUntrustedAggregateNamedChecks` covering both messages via log capture.

*This reply was generated by [Amp](https://ampcode.com) with Fable (claude-fable-5).*

docs/configuration.md
aparajon: Addressed in 6dc7d8c5 — the `TrustedCheckAppSlugs` field comment now describes the chain-level shared model (every deployment in a promotion chain carries the same complete list of the chain's App slugs, like `environment_order`) and both gates the list feeds: promotion-gate verification of a prior environment's aggregate, and passing-checks classification so a sibling's `action_required` aggregate is not treated as failing external CI. Configuring from the schema comment alone now produces a complete chain config.

*This reply was generated by [Amp](https://ampcode.com) with Fable (claude-fable-5).*

PR #313, opened 2026-06-11
feat(api): add live schema pull primitive

aparajon:
## Why
Schema onboarding needs a server-side primitive that can fetch live database state from the data plane before future CLI commands write declarative files.

## What
- Add a `PullSchema` Tern RPC and `/api/pull` control-plane endpoint
- Route pull requests through the existing local/gRPC client model
- Return MySQL schemas from live `SHOW CREATE TABLE` data with volatile counters stripped

## References
Payload shapes:

```json
POST /api/pull
{
  "database": "orders",
  "environment": "production",
  "type": "mysql"
}
```

```proto
message PullSchemaRequest {
  string database = 1;
  string type = 2;
  string environment = 3;
  string target = 4;
}

message PullSchemaResponse {
  string database = 1;
  string type = 2;
  string environment = 3;
  map<string, SchemaFiles> schema_files = 4;
  int32 table_count = 5;
}
```

`schema_files` is keyed by namespace, and each namespace contains `files` mapping filenames to declarative schema content.

Generated with Amp

### DIFF
--- a/pkg/api/handlers_test.go
+++ b/pkg/api/handlers_test.go
@@ -429,6 +429,9 @@ type mockTernClient struct {
 	planResp          *ternv1.PlanResponse
 	planErr           error
 	planReq           *ternv1.PlanRequest
+	pullSchemaResp    *ternv1.PullSchemaResponse
+	pullSchemaErr     error
+	pullSchemaReq     *ternv1.PullSchemaRequest
 	applyResp         *ternv1.ApplyResponse
 	applyErr          error
 	applyReq          *ternv1.ApplyRequest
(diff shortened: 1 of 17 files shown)


aparajon:
Triage review of the draft. The shape is right: engine-agnostic `map<string, SchemaFiles>` keeps engine specifics out of the API types, routing goes through the existing target resolution, and credentials stay server-side. The nil-proto-map comment is already addressed. Three things before marking ready:

1. **Validate pulled DDL is parseable at pull time** — the repo's contract is that all SQL must be TiDB-parseable, but pulled `CREATE TABLE` statements are returned unvalidated; an unparseable statement surfaces only later at plan time, far from the cause. Parse each table during pull and fail with the database/table name in the error.

2. **Multi-deployment environments get a generic 500** — the singular target resolution errors with "resolves to N deployments" for multi-deployment configs. Pull against an environment with multiple deployments needs either a clear 4xx with guidance or deliberate deployment selection in the request.

3. **gRPC error fidelity** — the server maps every failure to `codes.Internal`, so remote callers lose the unsupported-type/invalid-argument semantics the local path distinguishes. Map to `Unimplemented`/`InvalidArgument` where applicable so retry/UX classification works across the wire.

---

morgo (approved):

PR #316, opened 2026-06-12
feat(tern): add routing client

aparajon:
## Why
SchemaBot already has a routing contract that turns a logical database/environment request into concrete execution targets, but Tern clients still assume callers have already picked a specific deployment client. This adds the missing client seam so routing can be consumed centrally while preserving the route that was recorded when a plan or apply was created.

Before:

```text
API/caller
  │ picks deployment + target directly
  ▼
LocalClient / GRPCClient
  │
  ▼
execution target
```

After:

```text
API/caller
  │ logical database + environment
  ▼
routing.Resolver ── plan-time route
  │
  ▼
tern.RoutingClient
  │ delegates by stored route metadata
  ├─ Plan: current resolved target
  ├─ Apply: stored plan target
  ├─ Progress/control: stored apply target
  └─ Resume operation: stored operation target
      │
      ▼
LocalClient / GRPCClient
```

This is a good addition because route selection becomes explicit and durable at the client boundary: planning can use current routing, while apply/control/resume paths keep using the target that SchemaBot already stored for the work in flight.

## What
- Add `tern.RoutingClient` to resolve plan requests through `routing.Resolver` and delegate to deployment clients
- Route apply, progress, controls, and operation resume from stored plan/apply/operation metadata
- Preserve stored target values when delegating apply and operation-scoped resume requests
- Fail closed for multi-target plan routes until fanout is modeled end-to-end

## Risk Assessment
Low — this adds an unwired client wrapper and focused tests without changing existing server request routing.

Generated with Amp

### DIFF
--- a/pkg/tern/routing_client.go
+++ b/pkg/tern/routing_client.go
@@ -1,0 +1,444 @@
+package tern
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"google.golang.org/protobuf/proto"
+
+	ternv1 "github.com/block/schemabot/pkg/proto/ternv1"
+	"github.com/block/schemabot/pkg/routing"
+	"github.com/block/schemabot/pkg/storage"
+)
+
+// DeploymentClientFunc returns the Tern client that owns a deployment in an environment.
+type DeploymentClientFunc func(ctx context.Context, deployment, environment string) (Client, error)
+
+// PlanLookup loads stored plans by their external plan identifier.
+type PlanLookup interface {
+	Get(ctx context.Context, planIdentifier string) (*storage.Plan, error)
+}
+
+// ApplyLookup loads stored applies by their external apply identifier.
+type ApplyLookup interface {
+	GetByApplyIdentifier(ctx context.Context, applyIdentifier string) (*storage.Apply, error)
+}
+
+// ApplyOperationLookup loads operation-scoped routing metadata for a stored apply.
+type ApplyOperationLookup interface {
+	Get(ctx context.Context, id int64) (*storage.ApplyOperation, error)
+}
+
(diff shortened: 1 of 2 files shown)


aparajon:
Reviewed against the resolver seam from #311 — the wrapper consumes `routing.Resolver` cleanly, keeps control-plane gating out of the routed path, and the fail-closed posture (nil resolver, empty target, multi-target, ownership mismatch) is solid. Four items before wiring this in:

1. **`routing_client.go:152` — remote progress during the dispatch handoff window** — when the routed client is remote and the stored apply has no `ExternalID` yet, `ternApplyIdentifier` falls back to the local identifier and forwards it to a remote that can't know it. And since this wrapper hardcodes `IsRemote() == false`, the API layer's storage-serving guard for exactly this window is bypassed. Progress should fail closed (explicit error) until the external ID exists.

2. **`routing_client.go:178` — environment mismatch on control actions** — `Apply` rejects a request/stored environment mismatch, but Progress/Cutover/Stop/Start/Volume/Revert/SkipRevert silently overwrite `req.Environment` with the stored value. Control actions mutate state; a non-empty mismatched environment should be rejected the same way Apply does.

3. **`routing_client.go:342`** — `applyOperationID` is always 0 at every call site; suggest dropping it until a control path needs operation-scoped routing. Related: all controls route to `apply.Deployment` only — if operations ever span deployments, Stop/Cutover would reach only one. Worth failing closed or documenting the single-deployment assumption.

4. Nit: the pending observer is handed to the routed client before `client.Apply`; if Apply errors, it can attach to an unrelated later Apply on that client.

The two existing inline comments (observer attach on resume paths) look addressed on the current head.

---

morgo (approved):

PR #317, opened 2026-06-12
feat(config): add on_failure enum for rollout continuation

Kiran01bm:
## What and Why?

Introduces an `on_failure` enum (`halt | continue | pause`, default `halt`) as the per-environment rollout-continuation policy, replacing the `halt_on_failure` boolean knob. Behaviour matches the boolean for the two existing values; `pause` is parsed but validation-rejected until its release machinery lands.

on_failure: halt      ->  failed sibling blocks the rest of the rollout (default)
on_failure: continue  ->  terminal-failed sibling is exempted; rollout continues
on_failure: pause     ->  rejected in validation (follow-up)

- Config: `EnvironmentConfig.OnFailure` + `ServerConfig.OnFailure(database, environment)` resolver, defaulting to `halt`.
- Validation: requires a `deployments` map; rejects unknown values and `pause`.
- Storage: adds per-row column `on_failure varchar(16) NOT NULL DEFAULT 'halt'` (mirrors `cutover_policy`); picked up by `EnsureSchema`.
- Claim predicate: `FindNextApplyOperation` exempts a terminal-`failed` earlier sibling only when `on_failure = 'continue'`; every other value — including unrecognized ones — fails closed and keeps blocking.

Dormant until the multi-deployment fan-out is wired (one operation per apply today), so the field, column, and predicate are no-ops at runtime regardless of value.

### Schema rollout safety (two-step)

`apply_operations.halt_on_failure` already ships in the live schema, and `EnsureSchema` reconciles a renamed column by drop+add, not rename. Dropping it in the same change that adds `on_failure` would break still-running old pods mid-deploy, since they still query `halt_on_failure`. So this PR is the **additive** step — it adds `on_failure` and keeps `halt_on_failure`; a follow-up removes the old column once every pod runs the `on_failure` code.

Patch 1 (this PR):   add on_failure, keep halt_on_failure  -> old pods unaffected
Patch 2 (follow-up): drop halt_on_failure                  -> no code references it

The new `on_failure` column is additive (default `halt`), so no existing data is touched.

## Roadmap / upcoming PRs

This PR (the `on_failure` enum surface) is **merged**. The runtime continuation behaviour it gates lands in the B-track continuation-projection follow-ups; cross-referenced below so the enum surface and its consumers stay correlated.

| Upcoming change | Status | PR |
|---|---|---|
| `on_failure` enum surface (config/storage/predicate), additive column | ✅ Merged | #317 |
| Aggregate `applies.state` over all `apply_operations` rows ("Option B") | ✅ Merged | #318 |
| Policy-aware `continue` projection — hold apply active until siblings settle (**B2**) | Next | TBD |
| Eager failure surfacing + `stop` halts remaining siblings under `continue` (**B3**) | Planned | TBD |
| Drop `halt_on_failure` column (patch 2 above) | Planned | TBD |
| `on_failure: pause` as a release gate | Planned (depends on B2 + pause/release machinery) | TBD |
| Compile `on_failure` into dependency-graph edges (remove runtime knob) | Future (blocked on north-star Phases 4–5) | TBD |

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -467,15 +467,17 @@ type EnvironmentConfig struct {
 	// Only meaningful alongside a Deployments map.
 	CutoverPolicy string `yaml:"cutover_policy,omitempty"`
 
-	// HaltOnFailure controls multi-deployment rollout behaviour when a
-	// deployment fails. When true or unset (the default), a failed deployment
-	// halts the rollout — later deployments in deployment_order are not started.
-	// When false, a failed deployment no longer blocks later deployments, so
-	// the rollout attempts every deployment instead of stopping at the first
-	// failure. It governs only rollout continuation; the apply's pass/fail
-	// verdict and the merge gate stay fail-closed on any failed deployment.
-	// Only meaningful alongside a Deployments map.
-	HaltOnFailure *bool `yaml:"halt_on_failure,omitempty"`
+	// OnFailure controls multi-deployment rollout continuation when a deployment
+	// terminally fails. "halt" (the default, also used when unset) stops the
+	// rollout — later deployments in deployment_order are not started. "continue"
+	// drops a terminal-failed deployment as a blocker so the rollout attempts
+	// every deployment instead of stopping at the first failure. "pause" is a
+	// known value reserved for a future release-gate behaviour and is rejected
+	// at validation until that machinery lands. It governs only rollout
+	// continuation; the apply's pass/fail verdict and the merge gate stay
+	// fail-closed on any failed deployment. Only meaningful alongside a
+	// Deployments map.
+	OnFailure string `yaml:"on_failure,omitempty"`
 
 	// For PlanetScale/Vitess:
 	// Organization is the PlanetScale organization name.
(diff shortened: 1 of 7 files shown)


aparajon:
Reviewed for correctness and fail-closed behavior. The enum lands cleanly: `pause` parsed but validation-rejected, default `halt` enforced at resolver, insert fallback, and column default; the claim predicate exempts only exact `on_failure = 'continue'` with a `failed` earlier sibling, so unknown values fail closed. Three small items:

1. **`pkg/presentation/presentation.go:42`** — `Operation.HaltOnFailure`'s doc comment still describes dereferencing the stored `*bool`, which this PR deletes. Suggest updating the comment — or carrying the `on_failure` enum through the presentation input instead of a bool, so a future `pause` value doesn't need a second field re-plumb. (Note this field is also touched by #320; whichever merges second needs to absorb the change.)

2. **Fail-closed proof at the predicate** — could we add one claim-predicate test where a row carries an unrecognized `on_failure` value (e.g. `pause`) and assert the failed sibling still blocks? Config validation prevents it reaching storage today, but a storage-level test pins the safety property if validation ever changes.

3. **PR body nit** — the column replacement relies on EnsureSchema drop+add; worth a one-line mention that any environment that ran the prior column loses its values (fine since it never shipped, but good to state).

Verdict: LGTM with the above nits.

---

aparajon (approved):
Approving — fail-closed enum handling verified end to end. Nits in my earlier comment (stale presentation doc comment, predicate-level fail-closed test) are non-blocking.

Kiran01bm:
> * **`pkg/presentation/presentation.go:42`** — `Operation.HaltOnFailure`'s doc comment still describes dereferencing the stored `*bool`, which this PR deletes. Suggest updating the comment — or carrying the `on_failure` enum through the presentation input instead of a bool, so a future `pause` value doesn't need a second field re-plumb. (Note this field is also touched by [feat(github): dispatch apply comment on deployment count #320](https://github.com/block/schemabot/pull/320); whichever merges second needs to absorb the change.)
> * **Fail-closed proof at the predicate** — could we add one claim-predicate test where a row carries an unrecognized `on_failure` value (e.g. `pause`) and assert the failed sibling still blocks? Config validation prevents it reaching storage today, but a storage-level test pins the safety property if validation ever changes.
> * **PR body nit** — the column replacement relies on EnsureSchema drop+add; worth a one-line mention that any environment that ran the prior column loses its values (fine since it never shipped, but good to state).

Thanks for the note:

1.  presentation.go:42 doc comment → not actionable in #317 as `pkg/presentation/` doesn't exist on this branch and there are no `HaltOnFailure` references here as that field lives in #320 and since pull/317 removes the stored *bool, pull/320's stale doc comment is pull/320's fix to make when it rebases. Nothing to change here.
2. fail-closed proof at the predicate → done ✅ - added `UnrecognizedOnFailureBlocksFailedSibling` stores pause (an unrecognized value) on the failed sibling and asserts the rollout stays blocked, pinning the safety property even if config validation changes.
3. with the two-patch split, #317 no longer drops the column — on_failure is purely additive, so no data is lost. The drop is deferred to Patch 2.

PR #319, opened 2026-06-12
feat(github): render multi-deployment apply comment from presentation model

Kiran01bm:
## What and Why ?

Adds the GitHub markdown renderer for multi-deployment apply comments, consuming the surface-agnostic `presentation.Apply` model (from #315). This is the rendering half of MD-10 — it turns the derived presentation into the comment an operator sees on the PR, and seeds the multi-deployment scenarios into the preview pipeline so the UX is reviewable in `TEMPLATES.md`.

## Changes

- New renderer turns `presentation.Apply` into a multi-deployment apply comment:
  - aggregate header + state counts + next-action line
  - per-deployment summary list
  - a `<details>` block per deployment whose body reuses the existing single-deployment renderer, so per-deployment table fidelity is identical to today's single-deployment comment
- Seeds three multi-deployment apply preview scenarios — barrier in-progress, halt-on-failure, and completed rollout — through the preview pipeline so the aggregate + per-deployment UX auto-generates in `TEMPLATES.md`.
- This layer is pure rendering — no storage or dispatch wiring yet (that lands in the follow-up).

## Design-note compliance

Verified the seeded scenarios against the agreed UX contract in [`schemabot/design/schemabot-multi-deployment-ux.md`](https://github.com/squareup/schema-change-notes/blob/main/schemabot/design/schemabot-multi-deployment-ux.md):

| Design-note element | Covered? | Where |
|---|---|---|
| **Aggregate first, details second** | ✅ | All 3 seeds lead with aggregate header + `Deployments: …` counts, then per-deployment `<details>` |
| **Display order = `deployment_order` (eu→us→au→ca)** | ✅ | Rows and details emitted in that order in all 3 |
| **Open/collapsed defaults** (running/waiting/failed open; completed/pending collapsed) | ✅ | Barrier: eu+us `<details open>`, au/ca collapsed; Failure: us `<details open>`, eu/au/ca collapsed; Completed: all collapsed |
| **Failure surfaced at aggregate** (named failed dep + halted-because) | ✅ | Halt-on-failure seed: `❌ us — failed`, `⏸ au — halted — us failed` in the aggregate rows |
| **Single next action** | ✅ | One fenced command per scenario (cutover / retry) |
| **No internal terminology** | ✅ | "deployment", not `apply_operation` |
| **Invariant #9 — no parallel copy implied** | ✅ | Barrier seed has exactly **one** copy running (`us`); `eu` is parked at the cutover barrier — valid barrier semantics, not concurrent copies |
| **Future parallel-copy "multiple ready for cutover" example** | ⛔️ deliberately omitted | The note marks that as a not-yet-built future policy |

**Known divergence (follow-up):** the failure scenario's next-action renders `schemabot apply -e production`, whereas the design note's Failure-UX example prescribes a per-deployment `schemabot retry … --deployment <dep>` (with `revert` reserved for the revert-window only). That's existing renderer behavior, not seed data — tracked as a follow-up.

## Stacking

Part of the MD-10 per-deployment presentation series:
- #315 — derive surface-agnostic presentation model (merged)
- **this PR** — render multi-deployment comment from the model
- next — dispatch the apply comment on deployment count (single vs multi)

### DIFF
--- a/pkg/webhook/templates/multi_apply.go
+++ b/pkg/webhook/templates/multi_apply.go
@@ -1,0 +1,169 @@
+package templates
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/block/schemabot/pkg/presentation"
+)
+
+// MultiDeploymentApplyData is the input to the multi-deployment apply comment.
+// It bundles the surface-agnostic rollup (pkg/presentation) with each
+// deployment's existing single-deployment comment data, so the per-deployment
+// detail is rendered by exactly the same code path a single-deployment apply
+// uses today — the multi-deployment comment only adds the aggregate header and
+// the per-deployment hierarchy on top.
+type MultiDeploymentApplyData struct {
+	// Model is the derived rollup: aggregate state/label/counts/next-action and
+	// the per-deployment presentations in resolved deployment order.
+	Model presentation.Apply
+
+	// ApplyID is the parent apply's identifier, shown once in the aggregate header.
+	ApplyID string
+
+	// Environment is the rollout environment, shown in the header and used to
+	// format the next-action command.
+	Environment string
+
+	// RequestedBy is the operator who requested the apply.
+	RequestedBy string
+
+	// StartedAt / CompletedAt are RFC3339 timestamps for the aggregate elapsed time.
+	StartedAt   string
+	CompletedAt string
+
+	// Details maps a deployment name to that deployment's single-deployment
+	// comment data (its tables, error, timing, database). Each deployment's
+	// <details> body is rendered from its entry via RenderApplyStatusComment.
+	Details map[string]ApplyStatusCommentData
+}
+
+// RenderMultiDeploymentApplyComment renders the PR comment for an apply that
+// fans out across more than one deployment. The layout, per the multi-deployment
+// UX: an aggregate header (state title, metadata, per-status counts, and the
+// single next operator action), then a flat per-deployment summary so rollout
+// health and any failure are visible without expanding, then a <details> section
+// per deployment carrying today's per-table UX scoped to that deployment.
+//
+// The single-deployment case is intentionally not handled here: callers render
+// it with RenderApplyStatusComment so that UX stays byte-for-byte unchanged.
+func RenderMultiDeploymentApplyComment(data MultiDeploymentApplyData) string {
+	var sb strings.Builder
+
+	// Aggregate header: reuse the single-deployment title map keyed on the
+	// derived aggregate state so the headline vocabulary is identical.
+	writeApplyHeader(&sb, ApplyStatusCommentData{State: data.Model.State})
+	writeAggregateMetadata(&sb, data)
+	writeDeploymentCounts(&sb, data.Model.Counts)
+	writeAggregateNextAction(&sb, data)
+
+	// Flat per-deployment summary (always visible — survives any later size
+	// trimming of the detail sections).
+	writeDeploymentSummaryList(&sb, data.Model.Deployments)
+
+	// Expandable per-deployment detail, in resolved order.
+	writeDeploymentSections(&sb, data)
+
+	return sb.String()
+}
+
+// writeAggregateMetadata writes the apply-level metadata line. The database is
+// intentionally omitted — it is per-deployment and shown in each deployment's
+// section — so the aggregate carries only the environment, apply ID, elapsed
+// time, and requester.
+func writeAggregateMetadata(sb *strings.Builder, data MultiDeploymentApplyData) {
+	var parts []string
+	if data.Environment != "" {
+		parts = append(parts, fmt.Sprintf("**Environment**: `%s`", data.Environment))
+	}
+	if data.ApplyID != "" {
+		parts = append(parts, fmt.Sprintf("**Apply ID**: `%s`", data.ApplyID))
+	}
+	if elapsed := applyElapsed(ApplyStatusCommentData{StartedAt: data.StartedAt, CompletedAt: data.CompletedAt}); elapsed != "" {
+		parts = append(parts, fmt.Sprintf("**Elapsed**: %s", elapsed))
+	}
+	if len(parts) > 0 {
+		fmt.Fprintf(sb, "%s\n", strings.Join(parts, " | "))
+	}
+	writeAppliedByOrTimestamp(sb, data.RequestedBy)
+}
+
+// writeDeploymentCounts writes the per-status histogram so an operator sees
+// rollout health at a glance without expanding anything.
+func writeDeploymentCounts(sb *strings.Builder, counts []presentation.StateCount) {
+	if len(counts) == 0 {
+		return
+	}
+	parts := make([]string, 0, len(counts))
+	for _, c := range counts {
+		parts = append(parts, fmt.Sprintf("%d %s", c.Count, c.Label))
+	}
+	fmt.Fprintf(sb, "\n**Deployments**: %s\n", strings.Join(parts, ", "))
+}
+
+// writeAggregateNextAction renders the single suggested operator action derived
+// for the rollup, if any. The verb set mirrors the per-deployment commands the
+// CLI exposes; an empty action (NextActionNone) writes nothing.
+func writeAggregateNextAction(sb *strings.Builder, data MultiDeploymentApplyData) {
+	na := data.Model.NextAction
+	switch na.Kind {
+	case presentation.NextActionCutover:
+		writeFooterAction(sb,
+			fmt.Sprintf("To cut over `%s`:", na.Deployment),
+			fmt.Sprintf("schemabot cutover -e %s --deployment %s", data.Environment, na.Deployment))
+	case presentation.NextActionResume:
+		writeFooterAction(sb, "To resume:", fmt.Sprintf("schemabot start -e %s", data.Environment))
+	case presentation.NextActionReviewFailure:
+		if na.Deployment != "" {
+			writeFooterAction(sb,
+				fmt.Sprintf("Review the failure in `%s`, then revert:", na.Deployment),
+				fmt.Sprintf("schemabot revert -e %s --deployment %s", data.Environment, na.Deployment))
+		} else {
+			writeFooterAction(sb, "Review the failed deployment, then revert:",
+				fmt.Sprintf("schemabot revert -e %s", data.Environment))
+		}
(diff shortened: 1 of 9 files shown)


pkg/webhook/templates/multi_apply.go
```diff
@@ -0,0 +1,169 @@
...
+// section — so the aggregate carries only the environment, apply ID, elapsed
+// time, and requester.
+func writeAggregateMetadata(sb *strings.Builder, data MultiDeploymentApplyData) {
+	var parts []string
+	if data.Environment != "" {
+		parts = append(parts, fmt.Sprintf("**Environment**: `%s`", data.Environment))
+	}
+	if data.ApplyID != "" {
+		parts = append(parts, fmt.Sprintf("**Apply ID**: `%s`", data.ApplyID))
+	}
```
aparajon: Hmm this shouldn't ever be null, perhaps we can ensure this is validated upstream?

pkg/webhook/templates/multi_apply.go
```diff
@@ -0,0 +1,169 @@
...
+
+// writeAggregateMetadata writes the apply-level metadata line. The database is
+// intentionally omitted — it is per-deployment and shown in each deployment's
+// section — so the aggregate carries only the environment, apply ID, elapsed
+// time, and requester.
+func writeAggregateMetadata(sb *strings.Builder, data MultiDeploymentApplyData) {
+	var parts []string
+	if data.Environment != "" {
+		parts = append(parts, fmt.Sprintf("**Environment**: `%s`", data.Environment))
+	}
```
aparajon: Similar comment here, wondering if we can enforce this isn't null

pkg/webhook/templates/multi_apply.go
```diff
@@ -0,0 +1,169 @@
...
+		writeFooterAction(sb,
+			fmt.Sprintf("To cut over `%s`:", na.Deployment),
+			fmt.Sprintf("schemabot cutover -e %s --deployment %s", data.Environment, na.Deployment))
+	case presentation.NextActionResume:
+		writeFooterAction(sb, "To resume:", fmt.Sprintf("schemabot start -e %s", data.Environment))
+	case presentation.NextActionReviewFailure:
+		if na.Deployment != "" {
+			writeFooterAction(sb,
+				fmt.Sprintf("Review the failure in `%s`, then revert:", na.Deployment),
+				fmt.Sprintf("schemabot revert -e %s --deployment %s", data.Environment, na.Deployment))
```
aparajon: I think we might need more nuanced user guidance here for deployment failures. Some edge cases:

- `revert` only works for the `planetscale` engine currently, and only if the revert window is active (instant ddl doesn't go through the revert window)
- some failures could be transient failures that might exhaust the retry budget, but would succeed later on a retry
- some cases might warrant a `schemabot rollback` command, which should work across all engines
- some cases might warrant involving schemabot operators (e.g. the DB team)

aparajon:
The rendering matches the intended model well — aggregate header, per-status counts, single next-action, ordered summary, per-deployment `<details>` reusing the single-deployment renderer. Two correctness issues in the next-action commands need fixing, plus two smaller items:

1. **Rendered commands aren't executable by today's CLI** (`pkg/webhook/templates/multi_apply.go:117-129`) — `cutover`/`start`/`revert` all require an apply ID, and none of them have a `--deployment` flag yet. The single-deployment footer renders `schemabot cutover <apply-id>`. Suggest rendering the executable apply-ID form now, and switching to deployment-targeted commands in the PR that adds that CLI surface — the comment should never instruct an operator to run something that errors.

2. **`revert` is the wrong verb for a failed deployment** — revert targets a deployment inside its post-cutover revert window; a failed deployment never cut over. The existing failed-state footer suggests `schemabot apply -e <env>` to retry; the aggregate next-action should match that until a per-deployment retry command exists.

3. `<summary>` interpolates the deployment name/label as raw HTML without escaping — names come from config, so worth escaping defensively.

4. A deployment missing from `Details` renders an empty `<details>` body; a "no detail available" placeholder would read better.

There are also several existing inline review comments on these same lines that don't appear addressed on the current head — worth resolving those together with the above.

---

pkg/webhook/templates/multi_apply.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// deploymentTag renders the "<emoji> <deployment>" prefix, omitting the leading
+// space when a state has no glyph.
+func deploymentTag(d presentation.Deployment) string {
+	if d.Emoji == "" {
+		return d.Deployment
+	}
+	return fmt.Sprintf("%s %s", d.Emoji, d.Deployment)
+}
```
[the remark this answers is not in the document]
Kiran01bm: good point i'll escape html-escape them.

pkg/webhook/templates/multi_apply.go
Kiran01bm: good point thank you - tracing back confirms that the renderer guard is genuinely dead as the upstream boundary validation already exists at apply creation; 

Why ApplyID/Environment are guaranteed non-empty

```
apply creation (boundary)                renderer (#319)
─────────────────────────                ───────────────
apply_identifier = "apply-"+uuid   ┐
  plan_handlers.go:588             │  buildMultiApplyData (#320)
environment != "" enforced         │    ApplyID:     apply.ApplyIdentifier  ──▶  data.ApplyID
  api:   plan_handlers.go:92,324   ├──▶ Environment: apply.Environment       ──▶  data.Environment
  local: local_client.go:480       │    (verbatim pass-through)
NOT NULL columns                   │
  applies.sql:3 (apply_identifier) │
  applies.sql:10 (environment)     ┘
```

pkg/webhook/templates/multi_apply.go
Kiran01bm: yep agreed - i've responded the details here - https://github.com/block/schemabot/pull/319#discussion_r3407521825

Kiran01bm:
> * **Rendered commands aren't executable by today's CLI** (`pkg/webhook/templates/multi_apply.go:117-129`) — `cutover`/`start`/`revert` all require an apply ID, and none of them have a `--deployment` flag yet. The single-deployment footer renders `schemabot cutover <apply-id>`. Suggest rendering the executable apply-ID form now, and switching to deployment-targeted commands in the PR that adds that CLI surface — the comment should never instruct an operator to run something that errors.
> * **`revert` is the wrong verb for a failed deployment** — revert targets a deployment inside its post-cutover revert window; a failed deployment never cut over. The existing failed-state footer suggests `schemabot apply -e <env>` to retry; the aggregate next-action should match that until a per-deployment retry command exists.
> * `<summary>` interpolates the deployment name/label as raw HTML without escaping — names come from config, so worth escaping defensively.
> * A deployment missing from `Details` renders an empty `<details>` body; a "no detail available" placeholder would read better.

thank you - all 4 concerns are addressed in https://github.com/block/schemabot/pull/319/changes/51d578c70ad2e40c604e20f61445c53a2661846e

pkg/webhook/templates/multi_apply_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	assert.Contains(t, out, "## Schema Change In Progress")
+	assert.Contains(t, out, "**Environment**: `production`")
+	assert.Contains(t, out, "**Apply ID**: `apply-123`")
+	assert.Contains(t, out, "**Deployments**: 1 ready for cutover, 1 running, 2 waiting")
+
+	// Single next-action points at the cutover-ready deployment, even though the
+	// aggregate is still running.
+	assert.Contains(t, out, "To cut over `eu`:")
+	assert.Contains(t, out, "schemabot cutover -e production --deployment eu")
```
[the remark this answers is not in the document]
Kiran01bm: thank you - this is addressed in https://github.com/block/schemabot/commit/51d578c70ad2e40c604e20f61445c53a2661846e

pkg/webhook/templates/multi_apply.go
```diff
@@ -0,0 +1,169 @@
...
+// writeAggregateNextAction renders the single suggested operator action derived
+// for the rollup, if any. The verb set mirrors the per-deployment commands the
+// CLI exposes; an empty action (NextActionNone) writes nothing.
+func writeAggregateNextAction(sb *strings.Builder, data MultiDeploymentApplyData) {
+	na := data.Model.NextAction
+	switch na.Kind {
+	case presentation.NextActionCutover:
+		writeFooterAction(sb,
+			fmt.Sprintf("To cut over `%s`:", na.Deployment),
+			fmt.Sprintf("schemabot cutover -e %s --deployment %s", data.Environment, na.Deployment))
```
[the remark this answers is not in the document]
Kiran01bm: thank you - addressed in https://github.com/block/schemabot/commit/51d578c70ad2e40c604e20f61445c53a2661846e

pkg/webhook/templates/multi_apply.go
```diff
@@ -0,0 +1,169 @@
...
+	case presentation.NextActionResume:
+		writeFooterAction(sb, "To resume:", fmt.Sprintf("schemabot start -e %s", data.Environment))
```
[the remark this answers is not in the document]
Kiran01bm: thank you - addressed in https://github.com/block/schemabot/commit/51d578c70ad2e40c604e20f61445c53a2661846e

pkg/webhook/templates/multi_apply.go
Kiran01bm: thank you - fully-agree on all these observations - here's what i've got:

- Bullet 1 (revert is wrong for a failure): fully fixed. this was the concrete correctness bug — revert is invalid for a failed deployment (no cutover, no revert window, and it's PlanetScale-only) - removed the revert suggestion entirely and replaced it with the retry verb the single-deployment path already ships, so the rendered command is always executable today. 

- Bullets 2–4 (transient-retry nuance, cross-engine rollback, escalate-to-operators): not implemented — deferred. These require the renderer to branch on per-failure-type signals the presentation model doesn't expose yet (engine, revert-window eligibility, retry-budget state, fatal-vs-transient). Adding that means new NextAction kinds in the model (#315) plus the corresponding CLI verbs — which is the part of the plan for per-deployment CLI / manual-cutover slice (tracked as gap G7 in the derivation doc). Doing it in this presentation PR would be scope creep and would render verbs that don't exist in the CLI yet (the very thing we addressed in this issue comment).

aparajon (approved):

PR #322, opened 2026-06-12
feat(api): expose routing tern client

aparajon:
## Why
SchemaBot now has a generic routing contract and routing Tern client, but stock server construction still had no first-class way to assemble that client. Exposing this seam lets deployments opt into config-driven request routing without custom client wiring.

## What
- Add `Service.RoutingTernClient()` to build and cache a routing client from server config, storage, and existing deployment clients
- Route live schema pulls through `RoutingClient`, making `/api/pull` use the same target resolution path as the wrapper
- Cover service-level Plan/Apply assembly and live schema pull routing with focused tests

```diagram
Before:
╭──────────────╮     ╭──────────────────╮
│ custom code  │────▶│ deployment Tern  │
╰──────────────╯     ╰──────────────────╯

After:
╭──────────────╮     ╭──────────────────╮     ╭──────────────────╮
│ stock server │────▶│ RoutingTernClient│────▶│ deployment Tern  │
╰──────┬───────╯     ╰──────────────────╯     ╰──────────────────╯
       │
       ▼
╭──────────────╮
│ server config│
│ + storage    │
╰──────────────╯
```

## Risk Assessment
Low — routing client construction is opt-in, and the only handler moved onto it is the read-only live schema pull path. Apply/control handlers remain on deployment-scoped clients until transport-sensitive paths are handled deliberately.

Generated with Amp

### DIFF
--- a/pkg/api/handlers_test.go
+++ b/pkg/api/handlers_test.go
@@ -850,7 +850,10 @@ func TestExecutePullSchemaRoutesConfiguredMySQLTarget(t *testing.T) {
 		},
 	}
 	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelError}))
-	svc := New(&mockStorage{}, cfg, map[string]tern.Client{
+	svc := New(&mockStorageWithApplyStores{
+		plans:   &staticPlanStore{},
+		applies: &staticApplyStore{},
+	}, cfg, map[string]tern.Client{
 		"primary/production": mockClient,
 	}, logger)
 
--- a/pkg/api/plan_handlers.go
+++ b/pkg/api/plan_handlers.go
@@ -181,42 +181,42 @@ func (s *Service) ExecutePullSchema(ctx context.Context, req apitypes.PullSchema
 		return nil, unsupportedErr
 	}
 
-	client, err := s.TernClient(resolvedTarget.Deployment, req.Environment)
+	client, err := s.RoutingTernClient()
 	if err != nil {
 		span.RecordError(err)
 		span.SetStatus(otelcodes.Error, "tern client")
 		return nil, fmt.Errorf("database %q (%s): %w", req.Database, req.Environment, err)
 	}
 
-	s.logger.Info("ExecutePullSchema: calling client.PullSchema",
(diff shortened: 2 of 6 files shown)


aparajon:
Reviewed as the exposure seam for the routing client from #316. Good news first: all four deferred findings from the #316 review are addressed on main as merged — `ternApplyIdentifierForClient` now fails closed when a routed remote apply has no external ID, environment mismatches are rejected for all controls via `validateStoredApplyEnvironment`, operation-scoped routing landed (`ResumeApplyOperation` + `ApplyOperationLookup`), and the pending observer attaches only after plan routing resolves with cleanup on Apply error. This commit correctly completes the `tern.Client` interface (`PullSchema`) and the lazy accessor is properly memoized under `ternMu` and closed in `Service.Close`. Scope is right for a seam PR — no handler flips onto it yet.

Two items before merge, two smaller:

1. **Return the concrete `*tern.RoutingClient` from `RoutingTernClient()`, not `tern.Client`** — the interface return makes the accessor trivially substitutable into handler paths that branch on `IsRemote()` (the remote-progress-from-storage guard and several control handlers), and the wrapper's `IsRemote()` is still a hardcoded-`false` shim. Returning the concrete type forces a deliberate decision at each future call site, and the accessor's doc comment should carry the shim warning: safe for Plan/Apply/PullSchema routing, not yet for transport-sensitive paths.

2. **Service-level routed Apply test gap** — `service_test.go` covers Plan routing + memoization, but nothing exercises routed Apply (plan-time deployment/target pinning through the service accessor). Add at least a service-level routed Apply test here; the end-to-end integration test belongs in the PR that flips a handler onto the routing client.

3. No `slog.Info` when the routing client is lazily created, unlike the local tern client's creation log — lazy creation is a state transition on the request path; add the Info log for parity.

4. Add a `var _ Client = (*RoutingClient)(nil)` compile-time assertion (the gRPC client has one) — until this PR the routing client didn't satisfy `Client` at all, and the assertion makes interface completeness local and intentional.

One residual to carry into the wiring PR: two concurrent routed Applies against the same deployment client can swap/clear each other's pending observer — pre-existing in the deployment clients, but the routing layer makes it easier to hit.

---

morgo (approved):

PR #323, opened 2026-06-12
feat(cli): add `onboard` and `pull` commands

aparajon:
## Why
Make it easier to onboard an existing live MySQL database into SchemaBot's declarative schema layout without manually copying table definitions.

## What
- Add `schemabot onboard` to create `schemabot.yaml` plus namespace-scoped SQL files from live schema
- Add read-only `schemabot pull` to return the server-side pull payload without writing files
- Verify onboarded files produce no schema changes against the source environment by default
- Add dry-run previews, response validation, and overwrite protection with explicit `--force`

```text
schemabot onboard -d mydb -e production -s schema/mydb
        │
        ▼
SchemaBot API /api/pull ──▶ source environment live schema
        │
        ▼
schema/mydb/
├── schemabot.yaml
└── mydb/
    ├── users.sql
    └── orders.sql
        │
        ▼
plan verification against production: no schema changes
        │
        ▼
open a normal PR; SchemaBot reconciles other configured environments

schemabot pull -d mydb -e production
        │
        ▼
prints the /api/pull schema payload to stdout; no files are written
```

## Risk Assessment
Low — these are new opt-in CLI commands. `onboard` only writes local files, refuses to overwrite existing output unless explicitly forced, and validates the generated schema with a follow-up plan check; `pull` is read-only.

Generated with Amp

### DIFF
--- a/pkg/cmd/client/client.go
+++ b/pkg/cmd/client/client.go
@@ -52,6 +52,20 @@ func GetEnvironments(endpoint, database string) ([]string, error) {
 	return result.Environments, nil
 }
 
+// CallPullSchemaAPI fetches live schema files for a database/environment pair.
+func CallPullSchemaAPI(endpoint, database, dbType, environment string) (*apitypes.PullSchemaResponse, error) {
+	req := apitypes.PullSchemaRequest{
+		Database:    database,
+		Type:        dbType,
+		Environment: environment,
+	}
+	var result apitypes.PullSchemaResponse
+	if err := doPostInto(endpoint, "/api/pull", req, &result); err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
+
 // CallPlanAPI calls the plan API by reading .sql files from schemaDir.
 // Files are grouped by namespace: subdirectories become namespace keys,
 // flat files use the directory name as the namespace.
--- a/pkg/cmd/client/client_test.go
+++ b/pkg/cmd/client/client_test.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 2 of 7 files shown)


aparajon:
The core flow is in good shape: path-traversal validation on response namespaces/filenames, deterministic sorted output, fail-closed collision policy with exact colliding paths, MySQL-only gate, testify + `t.TempDir()` throughout, and the server-side pull API (#313) now validates pulled DDL and returns typed, actionable errors that this CLI surfaces correctly. Two items worth addressing before merge, plus smaller ones:

1. **No plan verification after writing files** — the command writes the pulled schema and `schemabot.yaml`, then prints "Next: open a normal PR" without confirming the written files produce a no-change plan against the live database. Without that, "onboard" is functionally just "pull," and a non-canonical export ships silently — the user finds out on their first real PR when the plan shows unexpected drift. Suggest calling the plan API for the pulled environment after writing and failing (or loudly warning) if the plan reports changes.

2. **`write()` is non-atomic with no partial-failure reporting** — an error mid-loop leaves a half-written schema directory, the error doesn't say which files were already written, and map-range iteration makes the partial set nondeterministic. Suggest iterating in the sorted `paths()` order and, on failure, reporting the files already written (or staging to a temp dir and moving into place).

3. An empty pull (0 tables, no namespaces) succeeds, writing only `schemabot.yaml` and printing "Pulled 0 tables" — almost certainly a wrong database/environment; fail with an actionable error instead.

4. `fileExists` conflates stat failure with not-exists, so a permission error during dry-run renders the path as new — `checkConflicts` already handles this properly; reuse that pattern.

5. The type check in `Run` is unreachable after the `enum:"mysql"` kong tag and repeats in `buildOnboardWritePlan` — keep the plan-level one (it validates the server response) and drop the duplicate.

6. Error UX: `plan` renders rich `APIError` output (status, error code) via a dedicated formatter; onboard returns a bare wrapped error. The multi-deployment and unsupported-type errors are exactly where the formatted output helps — consider reusing it.

7. Naming nits: `--force` vs `--overwrite` is worth a deliberate pick (overwrite describes the effect better), and the `pull` alias bakes in a second command name from day one — easy to add later, hard to remove.

---

morgo (approved):

PR #324, opened 2026-06-12
fix(github): support PR comment start command

aparajon:
Adds PR-comment support for `schemabot start <apply-id> -e <env>` by routing it through the same durable start control path used by the CLI/API.

Shares apply-scoped PR control validation across `stop`, `start`, and `cutover` so control commands consistently fail closed when the apply belongs to another PR/environment or actor authorization fails.

*This PR was generated by Amp.*

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -532,9 +532,9 @@ schemabot apply -e production
 | `schemabot apply -e <env>` | Plan, lock, and confirm deployment |
 | `schemabot apply-confirm -e <env>` | Execute a locked plan |
 | `schemabot unlock` | Release lock and discard plan |
-| `schemabot stop <apply-id>` | Stop an in-progress deployment |
-| `schemabot start <apply-id>` | Resume a stopped deployment |
-| `schemabot cutover <apply-id>` | Complete a deferred cutover |
+| `schemabot stop <apply-id> -e <env>` | Stop an in-progress deployment |
+| `schemabot start <apply-id> -e <env>` | Resume a stopped deployment |
+| `schemabot cutover <apply-id> -e <env>` | Complete a deferred cutover |
 | `schemabot rollback <apply-id> -e <env>` | Generate a rollback plan |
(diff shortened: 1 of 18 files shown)


aparajon:
Verified the properties that matter for this change: the start command is gated through the same actor-authorization path as stop/cutover before any side effect; `ExecuteStart` reuses the exact API start path (fresh apply re-resolve, `isStartRequestAllowedState` allowlist via `pkg/state`, pending-stop rejection, durable control-request queueing + operator wake for stopped applies), so the webhook never drives the engine directly for stopped work; and the shared `loadApplyForPRControl` keeps PR/environment scoping fail-closed across all three apply-scoped commands. TOCTOU is sound — ownership pre-checks use immutable fields and state gating happens on the fresh read inside `ExecuteStart`, so an apply that terminalizes between parse and dispatch falls out of the allowlist. Tests cover unauthorized, missing-apply-ID, durable-request recording, and idempotent duplicate requests.

A few items, none blocking:

1. `handleStartCommand` logs every `ExecuteStart` error as `Warn("start PR command rejected")`, but the error can be an internal failure (control store unavailable, storage error), not a rejection — split severity on `controlOperationHTTPStatus(err) >= 500` → `Error`. Stop/cutover share this pattern; fine as a follow-up covering all three.
2. Add a webhook test for the wrong-state path (start on a completed apply → actionable error comment) — that's the branch operators will hit most; the other rejection paths are covered.
3. `RenderStartCommandAccepted` compares the raw string `"already_requested"`, duplicating an unexported pkg/api constant — export a shared constant instead (stop has the same pre-existing issue).
4. `executeStartForApply` returns `http.StatusOK` paired with non-nil errors; the caller ignores status on error, but returning 0 (or documenting status as success-only) would prevent a future misuse.
5. The help table shows `schemabot start <apply-id>` but the command requires `-e <env>` — a copy-pasted bare command gets rejected. Worth showing the env flag in the stop/start/cutover rows.

Verdict: LGTM with the above nits.

---

morgo (approved):

PR #339, opened 2026-06-14
feat(tern): add static target router

aparajon:
## Why
A data-plane process needs a reusable way to serve many opaque execution targets without each embedder rebuilding inventory lookup and per-target client caching.

## What
- `tern.TargetRouter`, a `tern.Client` that resolves a target via `pkg/inventory` and caches a `LocalClient` **per target/type/environment** — a data plane has one deployment and many targets.
- Routes target-scoped pull, plan, apply, progress, controls, and operator resume through the cached clients.
- Plan routes by the resolved target; stored applies route by the target persisted on the apply, falling back to the plan-time target — so controls and resume reach the same data plane that ran the apply.
- The inventory DSN is namespace-free; the concrete MySQL schema is injected per Spirit operation by `LocalClient`, so one cached client per target serves every namespace.

```diagram
╭───────────────╮     ╭────────────────╮     ╭─────────────╮
│ Tern request  │────▶│ Target router  │────▶│ LocalClient │
│ target: key   │     │ key → client   │     │ per target  │
╰───────────────╯     ╰───────┬────────╯     ╰─────────────╯
                              │
                              ▼
                       ╭────────────────╮
                       │ inventory      │
                       │ key → DSN/meta │
                       ╰────────────────╯
```

## Risk
Low — additive and not wired into the default server path yet. Existing clients keep using the current routing unless a caller explicitly constructs the router. Builds on the merged `pkg/inventory` (#341) and namespace-free `LocalClient` execution (#340).

### DIFF
--- a/pkg/tern/local_client.go
+++ b/pkg/tern/local_client.go
@@ -744,6 +744,7 @@ func (c *LocalClient) Apply(ctx context.Context, req *ternv1.ApplyRequest) (*ter
 		DeferDeploy:  deferDeploy,
 		AllowUnsafe:  allowUnsafe,
 		SkipRevert:   skipRevert,
+		Target:       plan.Target,
 	}
 	optionsJSON := storage.MarshalApplyOptions(applyOpts)
 
--- a/pkg/tern/target_router.go
+++ b/pkg/tern/target_router.go
@@ -116,15 +116,21 @@ func (r *TargetRouter) Apply(ctx context.Context, req *ternv1.ApplyRequest) (*te
 	if req == nil {
 		return nil, fmt.Errorf("apply request is required")
 	}
-	target := targetOrDatabase(req.Target, req.Database)
+	// An apply executes a previously created plan, so the plan is authoritative
+	// for routing: it carries the execution target, namespace, type, and
+	// environment chosen at plan time. Derive the route from the plan and reject
+	// a request that contradicts it. Never treat req.Database as the target — the
+	// opaque execution target and the schema namespace are distinct, and that is
+	// the whole point of this router.
+	target := req.Target
 	if target == "" && req.Options != nil {
 		target = req.Options["target"]
 	}
 	namespace := req.Database
-	if namespace == "" && req.PlanId != "" {
-		if r.storage.Plans() == nil {
-			return nil, fmt.Errorf("plan lookup is required for target-routed apply %q", req.PlanId)
-		}
+	databaseType := req.Type
+	environment := req.Environment
+
+	if req.PlanId != "" && r.storage.Plans() != nil {
 		plan, err := r.storage.Plans().Get(ctx, req.PlanId)
 		if err != nil {
 			return nil, fmt.Errorf("load plan %q for apply target routing: %w", req.PlanId, err)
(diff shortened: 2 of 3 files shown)


aparajon:
I think this is the right direction for the data-plane routing seam. The current shape cleanly separates control-plane routing from data-plane target resolution: requests carry an opaque target, `TargetRouter` resolves that target through inventory, and the resolved connection config is delegated to a `LocalClient`. I also checked that the earlier cache/observer concerns are mostly addressed in the current head: cache keys include type/environment, pending observers can be request-scoped, `LocalClient.Apply` now persists `plan.Target`, and stored-apply routing can fall back to the stored plan target.

One thing I would fix before merging:

1. **Make `Apply` route by the stored plan target before falling back to request database.**

   `TargetRouter.Apply` currently initializes the routing target from `targetOrDatabase(req.Target, req.Database)`. If an apply request has `Database: "orders"` and no `Target`, that means the router tries to resolve `"orders"` and never consults the stored plan target. That conflates the schema namespace with the opaque execution target, which is exactly the distinction this router is trying to create.

   I think `Apply` should load the stored plan whenever `PlanId` is present and treat the plan as authoritative for routing:

   - `plan.Target` is the target to resolve
   - `plan.Database` is the namespace/database to pass to the local client
   - `plan.DatabaseType` is the database type
   - `plan.Environment` should be used, with a request-environment mismatch failing closed if the request supplies one

   A focused test would cover a request with `Database: "orders"`, empty `Target`, and a stored plan with `Target: "dsid-orders-prod"`; the resolver should only accept the plan target, and the delegated apply request should carry `Database: "orders"`, `Target: "dsid-orders-prod"`, and `Options["target"] == "dsid-orders-prod"`.

A couple of non-blocking notes:

- Including `database` in the cache key looks like the safer bridge for now. `LocalClient` still has namespace-bound behavior through `LocalConfig.Database`, so caching one client for all namespaces under the same target would be premature. It may be worth adjusting comments/body wording from “cached per target” to “cached per resolved target route and namespace” until `LocalClient` is fully namespace-agnostic.
- `clientForTarget` could defensively fail closed if the resolver returns nil or an incomplete result (`Target`, `DatabaseType`, or `DSN` empty), instead of assuming every resolver implementation is well-behaved.

Verdict: this is northstar-aligned and CI is green, but I’d fix the initial `Apply` routing issue before merging.

🤖 This review was generated by Amp with maintainer approval.

aparajon:
Thanks — correct catch, and fixed in `5394399f`.

`Apply` no longer derives its route from `targetOrDatabase(req.Target, req.Database)`. The plan is now authoritative for an apply's route: it loads the stored plan and takes the target, namespace, type, and environment from it, and **fails closed** when a request value contradicts the plan instead of silently overriding the plan-time route. It never falls back to the database name as the target — if no target resolves, it errors. This keeps the namespace/target distinction intact and honors "apply by the stored plan-time route."

I also added the symmetric guard your point implied: it's not only the empty-`Target` case — a *supplied-but-wrong* `Target`/`Database`/`Type`/`Environment` now fails closed against the plan, via a small `planAuthoritativeField` helper.

Tests:
- `TestTargetRouterApplyRoutesByPlanTargetWhenRequestOmitsTarget` — `Database:"orders"`, empty `Target`, stored plan `Target:"dsid-orders-prod"` → routes by the plan target; the delegated request carries `Database:"orders"`, `Target:"dsid-orders-prod"`, and `Options["target"]=="dsid-orders-prod"`.
- `TestTargetRouterApplyFailsClosedOnRequestTargetMismatch` — a request `Target` contradicting the plan fails closed with no client created.

Both non-blocking notes are also addressed in `48c9b5b8`: tightened the cache wording to "per resolved target route and namespace" (since `LocalClient` stays namespace-bound via `LocalConfig.Database`), and added a defensive fail-closed in `clientForTarget` for a nil/incomplete resolver result (with `TestTargetRouterFailsClosedOnIncompleteResolvedTarget`).

*🤖 Drafted by Claude Code.*

morgo (approved):

PR #348, opened 2026-06-15
feat(github): render multi-deployment terminal summary comment

Kiran01bm:
> Stacked on #346 — review/merge that first.

## What
Route the terminal **summary** comment through the same operation-count dispatch the status comment already uses. A fan-out apply (2+ operations) now gets an aggregate summary — terminal header, per-status counts, next action, and a `<details>` section per deployment carrying that deployment's own terminal summary. Single/zero-operation applies render the existing single-deployment summary byte-for-byte. Wired into both `OnTerminal` summary paths and the restart `ReconcileMissingSummaryComments` path.

## Why
After MD-10d wired the status comment for fan-out, the terminal summary was still single-deployment only — a multi-deployment apply would lose per-deployment outcomes in its final comment. An operation-load failure falls back to the single-deployment summary so a transient storage error never blocks the terminal comment.

```diagram
terminal apply ─▶ formatApplySummaryComment(apply, ops, tasks)
                        │
              len(ops)<=1 ┤ RenderApplySummaryComment            (unchanged)
              len(ops)>=2 ┘ RenderMultiDeploymentApplySummaryComment
                            (aggregate + per-deployment summaries)

Next
Terminal running_degraded surfacing and deployment-targeted next-action commands remain follow-ups.

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -4825,6 +4825,228 @@ ALTER TABLE `products` ADD INDEX `idx_price`(`price_cents`);
 
 
 </details>
+
+</details>
+
+<details>
+<summary><a name="summary-all-deployments-completed"></a><strong>Summary: All Deployments Completed</strong></summary>
+
+
+## ✅ Schema Change Applied
+
+**Environment**: `production` | **Apply ID**: `apply-a1b2c3d4e5f6` | **Elapsed**: 28m
+
+*Applied by @aparajon at 2026-01-01 00:00:00 UTC*
+
+**Deployments**: 3 completed
+
+- ✅ eu — completed
+- ✅ us — completed
+- ✅ au — completed
+
+<details>
+<summary>✅ eu — completed</summary>
+
+## ✅ Schema Change Applied
+
+**Database**: `payments_eu` | **Environment**: `production`
+
+
+> All 3 tables applied successfully — your schema changes are live!
+
+### ✅ testapp
+
+**`orders`**
+```sql
+ALTER TABLE `orders` ADD INDEX `idx_user_id`(`user_id`);
+```
+
+**`users`**
+```sql
+ALTER TABLE `users` ADD INDEX `idx_email`(`email`);
+```
+
+**`products`**
+```sql
+ALTER TABLE `products` ADD INDEX `idx_price`(`price_cents`);
+```
+
+_Apply ID: `apply-a1b2c3d4e5f6`_
+
+</details>
+
+<details>
+<summary>✅ us — completed</summary>
+
+## ✅ Schema Change Applied
+
+**Database**: `payments_us` | **Environment**: `production`
+
+
+> All 3 tables applied successfully — your schema changes are live!
+
+### ✅ testapp
+
+**`orders`**
+```sql
+ALTER TABLE `orders` ADD INDEX `idx_user_id`(`user_id`);
+```
+
+**`users`**
+```sql
+ALTER TABLE `users` ADD INDEX `idx_email`(`email`);
+```
+
+**`products`**
+```sql
+ALTER TABLE `products` ADD INDEX `idx_price`(`price_cents`);
+```
+
+_Apply ID: `apply-a1b2c3d4e5f6`_
+
+</details>
+
+<details>
+<summary>✅ au — completed</summary>
+
+## ✅ Schema Change Applied
+
+**Database**: `payments_au` | **Environment**: `production`
+
+
+> All 3 tables applied successfully — your schema changes are live!
+
+### ✅ testapp
+
+**`orders`**
+```sql
+ALTER TABLE `orders` ADD INDEX `idx_user_id`(`user_id`);
+```
+
+**`users`**
+```sql
+ALTER TABLE `users` ADD INDEX `idx_email`(`email`);
+```
+
+**`products`**
+```sql
+ALTER TABLE `products` ADD INDEX `idx_price`(`price_cents`);
+```
+
+_Apply ID: `apply-a1b2c3d4e5f6`_
+
+</details>
+
+</details>
+
+<details>
+<summary><a name="summary-halt-on-failure-one-deployment-failed"></a><strong>Summary: Halt On Failure (One Deployment Failed)</strong></summary>
+
+
+## ❌ Schema Change Failed
+
+**Environment**: `production` | **Apply ID**: `apply-a1b2c3d4e5f6` | **Elapsed**: 19m
+
+*Applied by @aparajon at 2026-01-01 00:00:00 UTC*
+
+**Deployments**: 1 completed, 2 halted, 1 failed
+
+---
+
+To retry:
+```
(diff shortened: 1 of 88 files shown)

### CODE REVIEW

**Comment by aparajon** [comment]
I think the terminal summary lifecycle still needs a multi-operation guard here.

This PR routes terminal summary rendering through the operation-count dispatch, but `OnTerminal` is still invoked by the tern poller when the current engine result is terminal. In the grouped path that can happen before the aggregate parent apply is terminal: one operation can finish while a sibling is still pending/running, especially under `on_failure=continue` or future fan-out. That would post/track the summary and clear the observer from the first terminal operation, not necessarily the final sibling.

Can we gate summary posting on the aggregate parent apply being terminal after deriving/reloading from all `apply_operations`? Otherwise a non-final sibling can publish the final summary early, and the restart reconciler will not repair it because a summary marker/comment already exists.

*This comment was generated by Amp.*

**Comment by aparajon** [comment]
The multi-deployment terminal summary should not depend on the `tasks` slice passed to `OnTerminal` unless that callback contract is changed to guarantee all parent-apply tasks.

In the operation-scoped operator path, the terminal callback may be reached while driving a single claimed operation, so the tasks passed into `formatTerminalSummaryComment` can be scoped to that operation rather than the whole apply. The new multi-summary renderer groups by `apply_operation_id`; with a partial task slice, sibling deployments can render without their table details even though those tasks are present in storage.

Can this path load all tasks for `apply.ID` before rendering a multi-deployment summary, or otherwise make the observer callback receive all parent tasks? The restart reconciliation path already does a full `Tasks().GetByApplyID`, but it only runs when the summary is missing, so it will not fix a partial summary that was already posted and tracked.

*This comment was generated by Amp.*

**Review by aparajon — APPROVED** [review]

**Comment by aparajon** [comment]
One more lifecycle concern: can we make the summary comment a **last-sibling-wins** side effect, not just rely on `apply_comments` upsert after posting?

`apply_comments` has a unique `(apply_id, comment_state)` row, but `postAndTrackComment` creates the GitHub comment before writing that row. If more than one worker ever reaches the terminal-summary path, both can create visible GitHub comments and the later upsert will only decide which one is tracked. That still leaves duplicate final summaries on the PR.

I think the desired invariant is stronger: the final summary should be emitted only by the worker that transitions the aggregate parent apply from non-terminal to terminal after all sibling `apply_operations` have settled. In other words:

1. a worker finishes and terminalizes its claimed operation row;
2. it re-derives the parent apply state from all sibling operation rows;
3. if the aggregate is still non-terminal, it does not post a summary;
4. if the aggregate becomes terminal, it wins a conditional parent-apply terminal transition;
5. only that winning worker posts the final summary.

That makes the last sibling / aggregate-terminal transition the single publisher, instead of letting multiple operation workers enter a GitHub side effect and using the DB row only as after-the-fact tracking. Open to a different implementation if it fits the composer northstar better; the important part is preserving the northstar shape where operation workers drive operation rows, the parent apply reflects the aggregate rollout, and the final PR summary is emitted once from that aggregate terminal transition.

*This comment was generated by Amp.*

**Comment by Kiran01bm** [comment]
> One more lifecycle concern: can we make the summary comment a **last-sibling-wins** side effect, not just rely on `apply_comments` upsert after posting?

thank you for all 3 comments and they share one root: `OnTerminal` fires per-driver on a terminal *engine result*, not once on the aggregate parent apply settling - this PR is rendering-only and stays correct while `1 apply == 1 apply_operation` (the multi path is never hit today), so I'd like to keep the lifecycle fix out of it.

also, the target invariant (which subsumes #1 and #2): the final summary is published only by the worker that wins a conditional non-terminal→terminal CAS on the parent apply after re-deriving it from all apply_operations — terminalize own op row → re-derive parent → if still non-terminal, don't post → if aggregate-terminal, win the CAS → only the winner loads all parent tasks (GetByApplyID) and posts exactly one summary - this makes the aggregate-terminal transition the single publisher instead of using `apply_comments` as after-the-fact tracking.

the work is tracked as a follow-up tied to the fan-out enablement (depends on native operation execution; overlaps #347/#349), so it can be tested end-to-end rather than bolted onto the renderer here.


### Before — per-driver OnTerminal (each terminal operation can publish)
```
                  ┌──────────────────────────────────────────────┐
                  │            apply (2+ operations)             │
                  └──────────────────────────────────────────────┘
        ┌──────────────────────┐            ┌──────────────────────┐
        │ worker A: op#1 drive │            │ worker B: op#2 drive │
        │  engine result=DONE  │            │  engine result=DONE  │
        └──────────┬───────────┘            └──────────┬───────────┘
                   │ OnTerminal(apply, tasks_op1)        │ OnTerminal(apply, tasks_op2)
                   │  (tasks = THIS op only)             │  (tasks = THIS op only)
                   ▼                                     ▼
        ╭──────────────────────╮            ╭──────────────────────╮
        │ no aggregate-terminal│            │ no aggregate-terminal│
        │ guard → proceeds     │            │ guard → proceeds     │  ◀── concern #1
        ╰──────────┬───────────╯            ╰──────────┬───────────╯
                   ▼                                     ▼
        format summary w/ PARTIAL tasks ◀── #2   format summary w/ PARTIAL tasks
                   │                                     │
                   ▼                                     ▼
   postAndTrackComment: create GH comment, THEN upsert apply_comments row
                   │                                     │
                   └──────────────┬──────────────────────┘
                                  ▼
              ╭───────────────────────────────────────────╮
              │  TWO visible GH summaries; upsert only    │
              │  decides which ROW is tracked   ◀── #3    │
              │  + observer cleared by first sibling      │
              ╰───────────────────────────────────────────╯
```

### After
```
        ┌──────────────────────┐            ┌──────────────────────┐
        │ worker A: op#1 drive │            │ worker B: op#2 drive │
        │  terminalize op#1 row│            │  terminalize op#2 row│
        └──────────┬───────────┘            └──────────┬───────────┘
                   ▼                                     ▼
        re-derive parent apply from ALL apply_operations rows
                   │                                     │
        ┌──────────┴───────────┐            ┌────────────┴─────────┐
        │ aggregate still       │           │ aggregate now        │
        │ non-terminal          │           │ TERMINAL             │
        │ (sibling pending)     │           │ (last sibling)       │
        └──────────┬───────────┘            └────────────┬─────────┘
                   ▼                                     ▼
            do NOT post  ◀── #1 fixed     Compare-And-Swap on applies row:
            (return)                      UPDATE ... SET state=terminal
                                          WHERE id=? AND state=non-terminal
                                                       │
                            ┌──────────────────────────┴───────────────────────┐
                            ▼                                                    ▼
                  rows_affected = 0                                   rows_affected = 1
                  (another worker already                            (THIS worker flipped
                   flipped it → lost)                                 it → won, happens once)
                            │                                                    │
                            ▼                                                    ▼
                      do NOT post                              load Tasks().GetByApplyID  ◀── #2 fixed
                                                               (ALL parent tasks)
                                                                         │
                                                                         ▼
                                                       post EXACTLY ONE summary  ◀── #3 fixed
```

PR #405, opened 2026-06-17
fix(api): filter Aurora/RDS-managed system schemas from pull discovery

aparajon:
## Why this matters

Aurora/RDS MySQL exposes several engine-managed system databases in `SHOW DATABASES` beyond the standard MySQL system schemas. The reserved-namespace filter that protects schema pull and onboarding didn't account for them, so onboarding an Aurora/RDS-backed database would discover these system schemas as user namespaces and try to materialize them into the declarative schema directory.

## What it does

Adds the Aurora/RDS-managed system databases — `innodb`, `tmp`, `polt`, `rdsmon`, `topo` — to `IsReservedPullNamespace`, alongside the existing reserved set (`information_schema`, `mysql`, `performance_schema`, `sys`, `dbadmin`, `schemabot`, `_pending_drops`, and `_`-prefixed schemas). This is the single filter point used by both live-namespace discovery and explicit pull-request rejection, so the fix covers both paths. Tests extended to cover the full system-schema set and case-insensitivity.

`PullSchema` is the routed live-schema primitive that onboarding, the Global Schema API, and future fleet scans all build on. Reserved-namespace filtering is what keeps that primitive returning only application namespaces across the fleet. Tightening it for Aurora/RDS deployments unblocks clean onboarding of those databases without leaking engine-managed schemas into generated files.


morgo (commented):
A little surprised by just "innodb", but no complaints excluding it.

pkg/schema/namespace.go
```diff
@@ -87,7 +87,7 @@ func GroupFilesByNamespace(files map[string]string, defaultNamespace string, env
 // from schema pull discovery and rejected for explicit pull requests.
 func IsReservedPullNamespace(namespace string) bool {
 	switch strings.ToLower(namespace) {
-	case "_pending_drops", "dbadmin", "information_schema", "mysql", "performance_schema", "schemabot", "sys":
+	case "_pending_drops", "dbadmin", "information_schema", "innodb", "mysql", "performance_schema", "schemabot", "sys", "tmp":
```
morgo: actually can you exclude polt, rdsmon and topo as well?

morgo (approved):

PR #431, opened 2026-06-19
feat(api): default operator claiming to the operation level

Kiran01bm:
## What

Make operation-level claiming the default for operator drivers. The
`operator_claim_operations` setting becomes a tri-state pointer:

- unset (default) → claim at the `apply_operations` level via `FindNextApplyOperation`
- explicit `true` → same operation-level path
- explicit `false` → fall back to apply-level claiming via `FindNextApply`

A new `ShouldClaimOperations()` accessor centralizes the nil-means-on logic.

## Why

The operation-level claim path is the foundation for multi-deployment applies.
While every apply still owns exactly one operation, the operation-scoped drive
resolves to the same work as the apply-scoped drive, so flipping the default is
behavior-preserving today while moving all new deployments onto the
multi-deployment-ready path without per-deployment config. The apply-level path
remains available via explicit `false` for fallback.

## Flow

Before — drivers default to apply-level claiming:

```
                operator_claim_operations unset
                              │
                              ▼
                    ┌──────────────────┐
   driver ─────────▶│  FindNextApply   │  (apply-level claim)
                    └────────┬─────────┘
                             ▼
                    drive whole apply
```
After — drivers default to operation-level claiming:
```
                operator_claim_operations unset  ──▶ ShouldClaimOperations() = true
                              │                           explicit false ─┐
                              ▼                                           ▼
                  ┌─────────────────────────┐            ┌──────────────────┐
   driver ───────▶│ FindNextApplyOperation  │            │  FindNextApply   │
                  └───────────┬─────────────┘            └────────┬─────────┘
                              ▼                                    ▼
                  drive per-deployment operation          drive whole apply
                  (multi-deployment ready)                  (fallback)
```
## References

Follow-up in the multi-deployment apply work stream; builds on the
operation-claim drive path landed in #429.


aparajon (commented):
Reviewed the default flip against the operator claim/lease path. Verified:
- `ShouldClaimOperations()` mirrors the existing `ShouldRequirePassingChecks()` nil-receiver/nil-field pattern and defaults on, with focused unit + YAML coverage for nil-receiver, nil-field, explicit true, explicit false, and both YAML cases.
- `new(false)`/`new(true)` are valid in the project's Go version and already used in the codebase.
- No tests were removed or weakened — the doubles that mock the apply-level `FindNextApply` claim path correctly pin `operator_claim_operations: false` (with comments explaining why), so they keep exercising the fallback path instead of being silently rerouted.
- Fail-closed safety holds end to end: invalid apply/operation lease, missing-deployment, and not-part-of-set branches all return without driving and emit a claim-failure metric; "parent not claimable" distinguishes terminal vs transient.
- The operation lease is attached to the drive context (operation/dual-lease contexts), so lease-guarded per-operation writes are authorized under the new default rather than bypassed.
- Mixed-version rollout is safe: apply-level and operation-level claims both contend on the parent apply row under `FOR UPDATE SKIP LOCKED` and rotate the same lease token, so a displaced driver's writes fail closed — no double-drive.

1. Minor doc wording: `ShouldClaimOperations` returns true to claim at the operation level; the doc comment reads slightly inverted ("returns whether ... claim at the apply_operations level"). Consider "returns true when drivers should claim at the apply_operations level."
2. The default changes on upgrade for any deployment that never set `operator_claim_operations`. It's behavior-preserving today (one operation per apply), but worth an explicit operator-facing note since it's a silent default change.

LGTM — both items are nits.

Kiran01bm:
> * Minor doc wording: `ShouldClaimOperations` returns true to claim at the operation level; the doc comment reads slightly inverted ("returns whether ... claim at the apply_operations level"). Consider "returns true when drivers should claim at the apply_operations level."
> * The default changes on upgrade for any deployment that never set `operator_claim_operations`. It's behavior-preserving today (one operation per apply), but worth an explicit operator-facing note since it's a silent default change.

thank you - both the nits are now addressed.

aparajon (approved):

PR #453, opened 2026-06-21
feat(tern): persist per-shard progress from the operator's lease-held drive

aparajon:
## Why this matters

#451 added the lease-held per-shard upsert; this calls it from the place that actually holds the operation lease — the operator's drive — so per-shard progress becomes durable and the renderer can show the per-shard breakdown from storage instead of a live `SHOW VITESS_MIGRATIONS` re-query.

## What it does

Wires `UpsertShardProgress` into `syncAtomicTaskProgress` (the grouped-apply progress sync). For each table with engine progress, it write-throughs `TableProgress.Shards` as per-shard `tasks`:

- **Lease-held single writer.** The write runs only inside the operator drive — `UpsertShardProgress` requires the operation lease, and `writeShardProgress` no-ops when no operation lease is on the context, so a read-path `Progress()` call never writes.
- **Expand phase only.** Per-table task rows are left untouched. Rendering and gating switch to the per-shard rows in a later change; the per-table anchor is removed for sharded engines only after that (the contract step).
- **Spirit unaffected.** Unsharded `TableProgress` carries no shards, so the write-through is a no-op for MySQL.

```
operator drive (holds the deploy-request operation lease)
  └── syncAtomicTaskProgress
        └── writeShardProgress → UpsertShardProgress (per shard)
```

Integration test: under the lease both shards persist with correct fields, a second pass updates in place (no duplicates), and without an operation lease the write-through is a no-op.

*This PR was written by Claude (Opus 4.8).*

### DIFF
--- a/pkg/storage/mysqlstore/tasks.go
+++ b/pkg/storage/mysqlstore/tasks.go
@@ -269,12 +269,15 @@ func (s *taskStore) insertShardTaskGuarded(ctx context.Context, task *storage.Ta
 	return nil
 }
 
-// GetByApplyID returns all tasks for an apply.
+// GetByApplyID returns the per-table tasks for an apply. Per-shard detail rows
+// (shard != "") are excluded: they are a reflected read-model written by the
+// operator and must not re-enter the per-table drive/gating/progress pipeline on
+// reload. Read those via GetShardProgressByApplyOperationID.
 func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage.Task, error) {
 	rows, err := s.db.QueryContext(ctx, `
 		SELECT `+taskColumns+`
 		FROM tasks
-		WHERE apply_id = ?
+		WHERE apply_id = ? AND shard = ''
 		ORDER BY created_at DESC
 	`, applyID)
 	if err != nil {
@@ -285,16 +288,17 @@ func (s *taskStore) GetByApplyID(ctx context.Context, applyID int64) ([]*storage
 	return scanTasks(rows)
 }
 
-// GetByApplyOperationID returns the tasks for a single apply_operation (one
-// deployment of a multi-deployment apply). While the config layer hard-blocks
-// more than one deployment per environment this returns the same rows as
-// GetByApplyID for the apply's single operation; it lets a driver drive and
+// GetByApplyOperationID returns the per-table tasks for a single apply_operation
+// (one deployment of a multi-deployment apply). It lets a driver drive and
 // reconcile one deployment independently once an apply fans out across several.
+// Per-shard detail rows (shard != "") are excluded for the same reason as
+// GetByApplyID: they must stay write-only relative to the per-table pipeline and
+// are read via GetShardProgressByApplyOperationID.
 func (s *taskStore) GetByApplyOperationID(ctx context.Context, applyOperationID int64) ([]*storage.Task, error) {
 	rows, err := s.db.QueryContext(ctx, `
 		SELECT `+taskColumns+`
 		FROM tasks
-		WHERE apply_operation_id = ?
+		WHERE apply_operation_id = ? AND shard = ''
 		ORDER BY created_at DESC, id DESC
 	`, applyOperationID)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


Kiran01bm:
issue surfaced from my agents review - 

CRITICAL ISSUE: persisted per-shard rows reprocessed as per-table tasks
═══════════════════════════════════════════════════════════════════════

```
DRIVE #1  (fresh apply — tasks set = anchors only, CLEAN)
╭──────────────────────────────────────────────────────────────────────╮
│ tasks = [ anchor(users, shard="") ]                                    │
│                                                                        │
│ syncAtomicTaskProgress(tasks, tp)                                      │
│   for anchor(users):                                                   │
│     engineProgressForTask → tp(users)        ── key=(ns,table) only    │
│     anchor.RowsCopied = tp.RowsCopied (table aggregate)  ✔ correct     │
│     writeShardProgress(anchor, tp) ─────────────┐                      │
│     transitionTaskState(anchor)                 │                      │
╰─────────────────────────────────────────────────┼──────────────────────╯
                                                  │ UpsertShardProgress
                                                  ▼ (per shard, lease-held)
                              ╭──────────────────────────────────────╮
                              │ STORAGE tasks                         │
                              │  anchor  users  shard=""              │
                              │  row     users  shard="-80"  ← NEW    │
                              │  row     users  shard="80-"  ← NEW    │
                              ╰──────────────────────────────────────╯
                                                  │
        operator re-claim / crash recovery /      │  GetByApplyID /
        stop-resume / any Progress() read         │  GetByApplyOperationID
                                                  ▼  (NO shard filter)
DRIVE #2  (RELOAD — tasks set now includes shard rows, BROKEN)
╭──────────────────────────────────────────────────────────────────────╮
│ tasks = [ anchor(shard=""), row(shard="-80"), row(shard="80-") ]       │
│                                                                        │
│ syncAtomicTaskProgress(tasks, tp)                                      │
│   for row(shard="-80"):                                                │
│     engineProgressForTask → tp(users)   ◀── matches (ns,table),        │
│                                             shard IGNORED              │
│   ┌─ (1) row.RowsCopied = tp.RowsCopied   �’️ TABLE aggregate written  │
│   │      onto a SHARD row  → corrupts per-shard data (in memory)       │
│   │                                                                    │
│   ├─ (2) writeShardProgress(row, tp)  → re-writes ALL shards again     │
│   │      O(shards²) writes per tick; briefly "repairs" -80 in storage  │
│   │                                                                    │
│   └─ (3) transitionTaskState(row)  → Update() persists the corrupted   │
│          aggregate from (1) back to STORAGE  ✗ re-corrupts shard -80   │
│      (order of rows decides which shard ends corrupted → racy)         │
│                                                                        │
│   ALSO: shard rows folded into                                         │
│     • deriveAggregateApplyState(taskStates(tasks))   → gating          │
│     • Progress() all-terminal gate + response table list → rendering   │
│   despite PR claiming "gating/rendering switch in a later change"      │
╰──────────────────────────────────────────────────────────────────────╯
```
CONSEQUENCES
  ✗ per-shard progress corrupted (the data this feature exists to store)
  ✗ O(shards²) write amplification every poll tick after first reload
  ✗ shard rows pollute aggregate state + all-terminal gate + Progress now
  ⚠ integration test never reloads → round-trip, so it cannot catch this

**Root cause (one line):**
The loaders (GetByApplyID / GetByApplyOperationID) and consumers (engineProgressForTask, syncAtomicTaskProgress, deriveAggregateApplyState, Progress()) have no notion of shard != "", so the feature's own persisted rows get pulled back into the per-table pipeline on the next reload.

Where the chain breaks (fix points): add a shard = '' filter (or skip shard != "" rows) at any one of — the drive/read loaders, or the top of syncAtomicTaskProgress/engineProgressForTask — so shard rows stay write-only until the deliberate contract step.

Kiran01bm (approved):

aparajon:
🤖 Great catch — confirmed as a real reload bug and fixed in 747dedcd.

Root cause was exactly as described: the persisted per-shard rows live in the same `tasks` table the per-table pipeline loads, but `GetByApplyID`/`GetByApplyOperationID` had no shard filter, so on the next drive/reclaim/progress read they were pulled back in and reprocessed as per-table tasks (table aggregate written onto a shard row, `writeShardProgress` re-running O(shards²), corrupted aggregate persisted, and shard rows leaking into aggregate state + the all-terminal gate that switch later).

Fix takes the loader-invariant approach you suggested:
- `GetByApplyID`/`GetByApplyOperationID` now return per-table tasks only (`shard = ''`), so the reflected shard rows stay write-only relative to the per-table pipeline until the deliberate render/gating switch.
- Added `GetShardProgressByApplyOperationID` to read the per-shard breakdown explicitly (for the renderer and tests).
- Tests now read shard rows through it and assert the per-table loader does **not** surface them — directly guarding the reload case the previous test missed.

Verified against the full `mysqlstore` + `tern` integration suites and the per-table pipeline (operator/progress/gating) integration tests.

PR #456, opened 2026-06-21
refactor(vitess): source deploy display metadata from the progress result

aparajon:
## Why this matters

`vitess_apply_data` is a side table that exists mostly so core can read deploy fields the engine already holds in its opaque `ResumeState.Metadata`. The progress API decodes those fields into the table, then reads them back to overlay onto the response. This is the first step of removing that table: have the engine surface its display fields directly, so nothing reads the side table for display.

## What it does

Adds a `Metadata map[string]string` to `engine.ProgressResult` and threads the PlanetScale deploy display fields (`branch_name`, `deploy_request_url`, `is_instant`, `deferred_deploy`) through it:

```
PlanetScale Progress() → ProgressResult.Metadata
  → ternv1.ProgressResponse.metadata (existing proto field)
  → progressResponseFromProto → apitypes ProgressResponse.Metadata
```

- The API's progress overlay **no longer reads `vitess_apply_data`** for those four display fields — they arrive on the response from the engine projection.
- `local_client` reads `is_instant` from the engine result instead of loading `vitess_apply_data`.

## Scope (expand phase)

`revert_skipped` is **core control state**, not engine data, so it still reads from the table here and relocates in the next change. The table's *writes* and the table itself are removed in the final step, once nothing reads it.

Tests: `psDisplayMetadata` field projection, and `progressResponseFromProto` carrying the metadata through. Build + vet + `engine`/`api`/`tern` unit suites green.

*This PR was written by Claude (Opus 4.8).*

### DIFF
--- a/pkg/api/handlers_test.go
+++ b/pkg/api/handlers_test.go
@@ -88,13 +88,21 @@ func (m *mockStorageWithPlanLookup) Plans() storage.PlanStore { return m.plans }
 
 type mockStorageWithApplyStores struct {
 	mockStorage
-	plans      storage.PlanStore
-	applies    storage.ApplyStore
-	tasks      storage.TaskStore
-	locks      storage.LockStore
-	applyLogs  storage.ApplyLogStore
-	controls   storage.ControlRequestStore
-	operations storage.ApplyOperationStore
+	plans           storage.PlanStore
+	applies         storage.ApplyStore
+	tasks           storage.TaskStore
+	locks           storage.LockStore
+	applyLogs       storage.ApplyLogStore
+	controls        storage.ControlRequestStore
+	operations      storage.ApplyOperationStore
+	vitessApplyData storage.VitessApplyDataStore
+}
+
+func (m *mockStorageWithApplyStores) VitessApplyData() storage.VitessApplyDataStore {
+	if m.vitessApplyData == nil {
+		return &staticVitessApplyDataStore{}
+	}
+	return m.vitessApplyData
 }
 
 func (m *mockStorageWithApplyStores) Plans() storage.PlanStore         { return m.plans }
@@ -114,8 +122,10 @@ func (m *mockStorageWithApplyStores) ApplyOperations() storage.ApplyOperationSto
 
 type staticApplyOperationStore struct {
 	storage.ApplyOperationStore
-	operations []*storage.ApplyOperation
-	err        error
+	operations      []*storage.ApplyOperation
+	err             error
+	resumeStateByOp map[int64]*storage.EngineResumeState
+	resumeStateErr  error
 }
 
 func (s *staticApplyOperationStore) ListByApply(_ context.Context, applyID int64) ([]*storage.ApplyOperation, error) {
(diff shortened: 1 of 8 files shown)


Kiran01bm:
from my agent's oracle

🔴 Critical: terminal/retryable/resuming applies lose branch_name / deploy_request_url / is_instant / deferred_deploy

The HTTP progress endpoint has two paths, and overlayVitessMetadata runs on both:
```
handleProgressByApplyID
  │
  ├─ shouldServeProgressFromStorage(state)?  ── terminal | failed_retryable | resuming
  │     └─ progressFromLocalStorage()         NO engine query
  │           └─ overlayVitessMetadata()  ── now sets ONLY revert_skipped
  │                                            ✗ branch_name / deploy_request_url /
  │                                              is_instant / deferred_deploy GONE
  │
  └─ active apply
        └─ client.Progress() → engine.Metadata → progressResponseFromProto
              └─ overlayVitessMetadata()  ── revert_skipped
                 ✔ display fields arrive from the engine projection
```
**Before this PR** `overlayVitessMetadata` read all five fields from vitess_apply_data regardless of path. Now the four display fields come only from engine.ProgressResult.Metadata, which is produced only when the engine is actually polled. But `progressFromLocalStorage` never calls the engine — and `shouldServeProgressFromStorage` routes `completed, failed, stopped, failed_retryable, and resuming` applies through it.

**Net effect:** for a completed PlanetScale apply (the common "let me look at what happened" case), the top-level Metadata map loses deploy_request_url (the link to the PlanetScale deploy request), branch_name, is_instant, and deferred_deploy. The PR description says "the API's progress overlay no longer reads vitess_apply_data for those four display fields — they arrive on the response from the engine projection," but that's only true on the live path. (Per-table IsInstant survives via `task.IsInstant` the top-level metadata keys do not.)

**Fix direction:** since the side-table is still written in this phase, keep progressFromLocalStorage's metadata sourcing intact — i.e. have overlayVitessMetadata still populate the four display fields from vad when serving from storage, or have progressFromLocalStorage read them. The clean removal of the storage read should wait until the engine projection is available on the storage path too (or the renderer is confirmed not to need these on terminal applies).

Kiran01bm (approved):

aparajon:
🤖 Good catch — confirmed, this is a real regression on the storage-served path.

`overlayVitessMetadata` runs on both paths, but the four display fields (`branch_name`, `deploy_request_url`, `is_instant`, `deferred_deploy`) now come only from the engine projection, which `progressFromLocalStorage` never produces. So `completed` / `stopped` / `failed_retryable` / `resuming` PlanetScale applies lose those top-level keys — and the CLI status box drops the `Branch` and `Deploy Request` rows, since `templates/progress.go` renders them unconditionally whenever present. That's exactly the "let me look at what happened" case.

Fix: while the side table is still written this phase, the storage path keeps sourcing those four fields from it — `overlayVitessMetadata` backfills them only when unset, so the live path stays engine-sourced and the storage path renders unchanged. Removing the storage read waits until the display fields are persisted into the read-model the storage path can read. Adding a test for the completed-apply-served-from-storage case. Thanks for the thorough trace.

aparajon:
🤖 Update — took a cleaner route than the backfill I described above. Those four display fields are already persisted durably in the operation's engine resume state (the same `psMetadataForStorage` the live path projects from), and that record is never deleted on completion. So the storage path (`progressFromLocalStorage`) now reads them straight from engine resume state — no `vitess_apply_data` read, no shim. Live-path values are preserved, and applies that predate resume-state persistence simply render without these fields.

Net effect: the side table is no longer read on either path here, so nothing blocks dropping it once `revert_skipped` moves to its own column. Pushed in b53e6508, with tests for the completed-apply-served-from-storage case (with and without resume state).

PR #481, opened 2026-06-23
fix(github): land a completed rollback's check as action_required

aparajon:
A rollback reverts a previously applied schema change, so once it completes the required status check must read `action_required` — the PR's change is gone from the environment and the PR must not merge as-is.

The rollback command registers an observer that does this, but operation-level operator claiming drives the rollback to terminal through the durable summary path, which suppresses that per-driver observer. That path was rollback-unaware and marked the check **successful**, so a completed PR-comment rollback left a green check that allowed the PR to merge with its change reverted.

This marks rollback applies durably (`storage.ApplyOptions.Rollback`) and teaches the durable terminal and stale-check reconciliation paths to finalize a completed rollback as `action_required`, matching the rollback command's own observer.

```
rollback-confirm → operator drives → terminal
  before:  refreshChecksForTerminalApply → success      ✗ PR mergeable, change reverted
  after:   refreshChecksForTerminalApply → action_required ✓ PR blocked until reconciled
```

### DIFF
--- a/pkg/webhook/check_records.go
+++ b/pkg/webhook/check_records.go
@@ -240,6 +240,16 @@ func (h *Handler) reconcileStaleChecks(ctx context.Context, client *ghclient.Ins
 			"apply_state", apply.State, "check_apply_id", check.ApplyID,
 			"check_head_sha", check.HeadSHA)
 
+		// A completed rollback must reconcile to action_required, not success: its
+		// success is a reverted PR change that must not merge as-is. Route it to the
+		// rollback finalizer instead of the ordinary apply-result update, which
+		// would mark the check successful.
+		if apply.IsRollback() && state.IsState(apply.State, state.Apply.Completed) {
+			h.setCheckActionRequired(repo, pr, apply.InstallationID, apply)
+			reconciled = true
+			continue
+		}
+
 		updated, err := h.updateCheckRecordForApplyResult(ctx, repo, pr, apply)
 		if err != nil {
 			metrics.RecordStatusCheckOperation(ctx, metrics.StatusCheckOperation{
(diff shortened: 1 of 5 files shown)


pkg/webhook/check_records.go
```diff
@@ -240,6 +240,24 @@ func (h *Handler) reconcileStaleChecks(ctx context.Context, client *ghclient.Ins
 			"apply_state", apply.State, "check_apply_id", check.ApplyID,
 			"check_head_sha", check.HeadSHA)
 
+		// A completed rollback must reconcile to action_required, not success: its
+		// success is a reverted PR change that must not merge as-is. Route it to the
+		// rollback finalizer instead of the ordinary apply-result update, which
+		// would mark the check successful.
+		if apply.IsRollback() && state.IsState(apply.State, state.Apply.Completed) {
```
Kiran01bm, line 247: Minor observability inconsistency - flagged by my agent -
- the rollback branch in `reconcileStaleChecks` unconditionally sets `reconciled = true` and records a "success" `stale_check_reconciliation` metric, even if s`etCheckActionRequired `internally skipped on an ownership miss (it returns void). 
- the non-rollback branch only counts success when updated. `setCheckActionRequired` still emits its own skipped/ownership-miss metric, so signal isn't lost — but the stale-reconcile success count can slightly overcount, and it triggers an aggregate refresh on a no-op.

Kiran01bm (approved):

pkg/webhook/check_records.go
aparajon: 🤖 Good catch — fixed in 6977cf47. `setCheckActionRequired` now returns whether the stored check was actually updated, and the reconcile fast-path only records the `stale_check_reconciliation` success (and sets `reconciled`) when it did. An ownership-miss skip no longer inflates the success count or triggers a spurious post-loop aggregate refresh; the skip path still emits its own `rollback_finished` skipped metric, so no signal is lost.

PR #488, opened 2026-06-23
feat(storage): add apply operation keys

aparajon:
## Why
Allow apply operation storage to distinguish multiple execution operations that share the same apply and deployment.

## What
- Add `operation_key` to the `apply_operations` storage identity
- Preserve legacy unkeyed lookup behavior for existing single-operation callers
- Add keyed lookup and storage coverage for uniqueness and round-tripping

## Risk Assessment
Low — this is a storage identity extension with a backward-compatible empty-key default for existing callers.

Generated with Amp

### DIFF
--- a/pkg/schema/mysql/apply_operations.sql
+++ b/pkg/schema/mysql/apply_operations.sql
@@ -2,6 +2,7 @@ CREATE TABLE `apply_operations` (
   `id` bigint unsigned NOT NULL AUTO_INCREMENT,
   `apply_id` bigint unsigned NOT NULL,
   `deployment` varchar(255) NOT NULL,
+  `operation_key` varchar(255) NOT NULL DEFAULT '',
   `target` varchar(255) NOT NULL DEFAULT '',
   `engine_resume_context` varchar(255) DEFAULT NULL,
   `engine_resume_metadata` json DEFAULT NULL,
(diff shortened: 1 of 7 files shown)


aparajon:
Reviewed the storage plumbing, the migration semantics, and callers.

**Verified sound:** the INSERT column list, placeholders, and arg list all line up (12/12/12), and `scanApplyOperationInto` reads `operation_key` in the same ordinal position as `applyOperationColumns` — no off-by-one. Backward compatibility holds: callers that don't set a key get the `NOT NULL DEFAULT ''` column, and the widened unique key `(apply_id, deployment, operation_key)` still enforces the old `(apply_id, deployment)` guarantee for them (always `''`), so it's safe under a rolling deploy and a rollback. The index change is a relaxation, and every existing row maps to `operation_key=''` (which already satisfied the 2-column uniqueness), so the rebuild can't fail on existing data. `GetByApplyAndDeployment` has no non-test callers today, so delegating it to the empty-key lookup is a no-op for current code while `GetByApplyDeploymentAndOperationKey` becomes the forward path. Tests cover dup-key, distinct-key, distinct-deployment, and multiple-keys-per-deployment with specific assertions, and none were deleted. Removing the stale "no caller reads/writes these rows yet" note is correct.

Two small things, neither blocking:

1. The `ApplyOperation` doc comment describes the slice as `(deployment, operation_key, target)`, but `target` isn't part of the storage identity. `deployment` → `target` is a strict 1:1 resolution (the config resolver maps each deployment key to exactly one target), so `target` is a *derived attribute* of `deployment`, not an independent identity dimension. The unique key `(apply_id, deployment, operation_key)` is correct as-is — including `target` in it would be redundant and would fork identity if a deployment ever re-resolved to a new target. Suggest updating the comment to `(deployment, operation_key)` and documenting `target` as the resolved address of the deployment. (This is also why no `target` dimension is needed for multi-region fan-out: the region *is* the `deployment` dimension, with `operation_key` distinguishing operations within it.)

2. The `ListByApply` comment now says "stable claim order"; since the query is `ORDER BY created_at, id` and the tests assert exact order, stating `(created_at, id)` keeps the contract precise.

On Copilot's `operator` → `driver` suggestion: I'd keep `operator` — the claim loop is a real named component, and the comments are consistent with the surrounding code; the driver-terminology guideline is about avoiding "worker." Your call.

Verdict: LGTM with the two minor comment nits. The storage identity and migration are sound.

morgo (approved):

PR #489, opened 2026-06-23
feat(storage): persist shard plan metadata

aparajon:
## Why
Apply creation needs durable shard membership and drift from the original plan request so later execution can reconstruct per-shard operation groups from stored data.

## What
- Add generic shard plan metadata to stored plans
- Persist shard metadata from Tern plan responses in existing plan data JSON
- Add storage and API coverage for shard metadata conversion and round-tripping

## Risk Assessment
Low — this extends stored plan metadata without changing the database schema or execution behavior.

Generated with Amp


aparajon:
Reviewed the diff and edge cases; built, vetted, and ran the new unit tests against the PR head (all green; the round-trip test is integration/DB-gated).

**Verified sound:** the no-DDL approach is correct — shard metadata rides in the existing `plan_data` JSON via `NamespacePlanData.Shards`. Backward/forward compatible: old `plan_data` without a `shards` key decodes cleanly (`TestPlanStore_LoadsPlansWithoutShardPlans`), old binaries ignore the new optional field, and plans are immutable so nothing re-writes and drops it — no migration needed. `ShardPlan{Shard, Namespace, NeedsChange}` is generic/engine-agnostic. Ordering is deterministic (sorted by `(namespace, shard)` on both write and read), so reconstruction is stable. `storePlanResponse` now copies `resp.Shards` with nil-rejection in the converter. Tests cover the API capture path, the converter (incl. nil), and a full store round-trip with specific assertions; none deleted.

Two small things, neither blocking:

1. `namespacesWithShardPlans` makes the flattened `plan.Shards` authoritative when set (`if len(plan.Shards) > 0 { shards = nil }`). If a caller ever sets *both* `plan.Shards` and per-namespace `nsData.Shards` for different namespaces, the per-namespace ones are silently dropped. Current callers set only one, so it's fine today — worth a one-line doc comment stating the precedence so it doesn't become a future footgun.

2. Minor round-trip change: `Create` previously marshaled `plan.Namespaces` directly (nil → `null`), but now marshals `namespacesWithShardPlans(plan)`, which returns a non-nil empty map for nil input (→ `{}`). Harmless for `len()`/`range` consumers and unasserted by tests, but if you want byte-identical output for the degenerate nil case, short-circuit to `nil` when both `Namespaces` and `Shards` are empty.

Nit: the `Plan.Shards` doc comment blends two ideas — *where* the data lives (`Namespaces[].Shards` is the source of truth) and *why* old rows still decode (the field is optional). Could split for clarity.

Verdict: LGTM with the doc/edge-case nits above. The storage shape and round-trip are sound, and this is independent of the operation-key change, so it can land in either order.

morgo (approved):

PR #511, opened 2026-06-25
fix(github): restore VSchema application status in progress views

aparajon:
## Why this matters

VSchema application status had drifted off operator-facing progress. #506 moved it onto engine display metadata and #507 removed the synthetic `vschema_update` task — but the GitHub PR comment was never wired to the metadata, so VSchema status silently disappeared from PR comments, and the CLI preview scenarios still modeled the removed task (rendering VSchema as a generic progress bar + "Running…").

A VSchema change can also span multiple keyspaces that apply independently, so a single combined status/diff blob can't represent it faithfully.

## What it does

- **PR comment** — renders a `### VSchema` section from the engine's display metadata, the same source the CLI uses. A VSchema-only apply (no per-table tasks) shows just this section, and its completed summary reports the VSchema outcome instead of "All 0 tables applied".
- **Per keyspace** — VSchema is modeled as a list of `{namespace, status, diff}` entries, so a multi-keyspace deploy renders each keyspace independently on both surfaces.
- **One source of truth** — the entries ride in display metadata as JSON under a shared key; the engine encodes, both renderers decode via `apitypes.ParseVSchemaChanges`, and the status label is shared via `ui.VSchemaStatusLabel`, so the CLI and PR comment can't drift apart again.
- **Cleanup** — removes the dead synthetic-`VSchema:`-table handling in the progress renderer and corrects a stale engine doc comment.

```
engine resume metadata
   └─ vschema_changes: [{namespace, status, diff}, …]   (JSON in display metadata)
        │
        ├─► PR comment : "### VSchema" → per-keyspace status + diff block
        └─► CLI / watch : "~ VSchema (<keyspace>): <status>" + diff
```

PlanetScale has a single deploy-level VSchema phase today, so every keyspace carries the same status. The per-keyspace shape is forward-looking for Strata, where keyspaces apply VSchema independently — that engine can populate per-keyspace statuses with no further rendering change.

<details>
<summary>Preview: multi-keyspace VSchema in an in-progress PR comment</summary>

## Schema Change In Progress

**Database**: `testapp` | **Environment**: `staging` | **Apply ID**: `apply-a1b2c3d4e5f6` | **Elapsed**: 8m

### VSchema

**`commerce`**: Applied

```diff
+ "lookup_orders": {"type": "lookup_hash"}
```

**`commerce_sharded`**: Applying...

```diff
+ "xxhash": {"type": "xxhash"}
```

</details>


Kiran01bm:
from my agent

**Critical:** the ### VSchema section never renders on real PR comments

The PR's headline fix — "PR comment renders a ### VSchema section from the engine's display metadata" — is not wired into the production webhook path. The new field ApplyStatusCommentData.VSchemaChanges is only ever populated by ApplyStatusFromProgress, and that function has zero production callers (only its definition + tests; git grep ApplyStatusFromProgress across the tree confirms).

Every live PR comment is built by buildApplyCommentData(apply, tasks), which never sets VSchemaChanges:

CommentObserver.formatStatusComment / formatTerminalSummaryComment   (pkg/webhook/comment_observer.go)
  └─ formatApplyStatusComment / formatApplySummaryComment            (single op today → len(ops) <= 1)
       └─ RenderApply{Status,Summary}Comment( buildApplyCommentData(apply, tasks) )
                                               └─ sets ApplyID/Database/State/Tables… but NOT VSchemaChanges

buildApplyCommentData ([pkg/webhook/apply.go](https://github.com/block/schemabot/blob/aparajon/vschema-render-fix/pkg/webhook/apply.go)) reads only apply + tasks (storage) — it never touches the engine display metadata where vschema_changes lives. The multi-deployment path has the same gap (buildMultiApplyData → buildDeploymentDetail also doesn't set it).

**Impact:**
- writeVSchemaStatus always returns early (len(data.VSchemaChanges) == 0) → the ### VSchema section is invisible on real PR comments. The bug the PR set out to fix (#506/#507 dropped VSchema from comments) is not actually fixed.
- A real VSchema-only completed apply hits case totalTables == 0: → "Schema change applied successfully", so the intended case totalTables == 0 && len(VSchemaChanges) > 0: → "VSchema applied successfully" branch is dead code in production.
- The tests pass only because they bypass buildApplyCommentData — they call ApplyStatusFromProgress or construct ApplyStatusCommentData{VSchemaChanges: …} directly. Classic green-tests / broken-prod-path gap.

Contrast — the CLI side is correctly wired: progress_parse.go builds ProgressData{… Metadata: result.Metadata}, and WriteProgress decodes via apitypes.ParseVSchemaChanges(data.Metadata). So the "one source of truth, can't drift" claim is half-true: the CLI reads the source; the PR comment doesn't read it at all.

**Fix direction:** populate VSchemaChanges inside buildApplyCommentData (and buildDeploymentDetail). Note this isn't a one-liner — that builder only has apply + tasks; the vschema_changes blob is produced by psDisplayMetadata during Progress and lives on the apply's engine resume state, so the observer needs to load/project that metadata (or call ApplyStatusFromProgress from a progress response) before rendering. Add a test that drives the real comment_observer → buildApplyCommentData path, not just the template functions.

**Secondary / worth confirming**

- Empty in-progress comment for VSchema-only applies (webhook): once tables are gone (post-#507) and VSchemaChanges is empty (this bug), an in-progress VSchema-only apply renders a body with no progress section and no VSchema section. After fixing the wiring this resolves, but worth a preview/test.
- nextVSchemaStatus vs. the per-keyspace label: status only reaches "applied" if a poll observed in_progress_vschema (current == "applying"). If a fast/instant VSchema-only deploy completes between polls, status stays "" → VSchemaStatusLabel("") = "Pending". Once the comment wiring is fixed, the completed summary could then show **ks**: Pending under a "VSchema applied successfully" header. This is pre-existing in the status machine but newly visible via per-keyspace rendering — confirm there's a terminal reconciliation that forces applied on completion.
- Behavior change (benign): totalTables == 0 previously rendered "All 0 tables applied successfully"; now "Schema change applied successfully". That's an improvement, just call it out as an intentional copy change.

The structural refactor itself (per-keyspace VSchemaChange, shared ui.VSchemaStatusLabel, apitypes encode/decode, engine VSchemaDiff→VSchemaDiffs) is clean and correct. The blocker is purely that the producer side (buildApplyCommentData) was never connected to the new consumer.

Kiran01bm (approved):

aparajon:
🤖 _(replying via Armand's AI agent)_

Great catch — that was a real green-tests/broken-prod gap. Fixed end to end:

- **Wiring**: `buildApplyCommentData` / `buildDeploymentDetail` now populate `VSchemaChanges`, projected per-operation from the apply's engine resume state via `tern.PSDisplayMetadata` — the same storage-backed projection `overlayStoredDisplayMetadata` uses. Resolved at the observer and handler; the rollback initial comment passes nil and the observer refreshes it on the next tick.
- **Second gap underneath it**: the stored-state projection (`tern.PSDisplayMetadata`) is a separate copy that was still emitting the old `vschema_status`/`vschema_diff` and decoding the old field. Updated it to decode `vschema_diffs` and emit the per-keyspace `vschema_changes` blob, so the API stored path and the comment path agree.
- **Status reconciliation** (your secondary point): `nextVSchemaStatus` now forces `applied` on deploy completion, so a fast VSchema-only deploy that passes between polls never renders "Pending" under an "applied successfully" header.
- **"All 0 tables"**: the VSchema-only completed summary now reports the VSchema outcome instead.
- **Coverage**: added `TestCommentObserverRendersVSchemaFromEngineResumeState`, which drives the real observer → `buildApplyCommentData` path from a stubbed engine resume state — progress comment, terminal summary (VSchema-only, asserts no "0 tables"), multi-keyspace, and the fail-soft cases (missing resume state / non-PlanetScale).

Commits 201df14 (wiring) + 0ab9c63 (coverage)

PR #531, opened 2026-06-25
feat(tern): fan sharded applies out per shard on the data plane

morgo:
A data-plane Tern drives applies inline via LocalClient.Apply, which built a single apply_operation carrying namespace-level tasks with no shard. A sharded engine (e.g. GAP's Strata engine) requires exactly one target shard per work operation, so it rejected these applies with "expected exactly one target shard, got 0" — the sharded plan was never fanned out, because the fan-out only existed in the control-plane api.Service path and is gated on a client capability the GRPCClient does not implement.

Teach LocalClient.Apply to fan out when the operator claims at the operation level (LocalConfig.ClaimOperations, mirrored from ShouldClaimOperations):

- buildApplyOperationGroups builds one work operation per (namespace, shard, table) with a single shard-tagged task for a sharded engine whose plan carries changing shards, plus a task-less group_finalizer per VSchema- changing namespace; any other apply stays a single whole-deployment work operation. This mirrors the control-plane fan-out for one deployment.
- In ClaimOperations mode Apply creates the operations and wakes the operator to claim and drive each (handing the engine one shard per operation) instead of driving the whole apply inline; mixing inline drive with operation claiming is unsafe because the inline path holds the apply lease but not the operation lease. The legacy single-operation, inline-driven path is unchanged when ClaimOperations is false (the default).

Thread ClaimOperations from ServerConfig.ShouldClaimOperations() at the LocalClient construction sites (serve grpcLocalClientFactory, api.Service).

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -551,6 +551,7 @@ func (s *Service) newLocalTernClient(key, database, dbType string, envConfig Env
 		Metadata:        metadata,
 		WakeOperator:    s.wakeOperator,
 		EngineFactories: s.engineFactories,
+		ClaimOperations: s.config.ShouldClaimOperations(),
 	}, s.storage, s.logger)
 	if err != nil {
 		return nil, fmt.Errorf("create local tern client for %s: %w", key, err)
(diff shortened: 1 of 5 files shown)


morgo:
🤖 Closing in favor of the control-plane fan-out approach. Fanning out on the data plane (this PR) prevents the control plane from sequencing per-shard rollout (canaries) and reconciling per-shard drift, and it reimplements the fan-out incompletely (see the two review comments — both are cases the control-plane `buildShardedApplyOperationGroups` already handles). The correct fix: `GRPCClient` advertises `ShardedApplyFanoutSupport` so the control plane fans out, and the data-plane `LocalClient.Apply` honors the per-shard scoped dispatch (`TargetShards`/`DdlChanges`). Will open a replacement PR.

morgo:
🤖 Replacement PR (control-plane fan-out): https://github.com/block/schemabot/pull/535

PR #548, opened 2026-06-26
fix(tern): avoid nil deref when engine Progress errors in conflict check

morgo:
## Problem

An Apply RPC panicked with a nil pointer dereference, surfaced by the grpc recovery interceptor as `Internal desc = runtime error: invalid memory address or nil pointer dereference`:

```
tern.(*LocalClient).tryResolveStaleTask  .../pkg/tern/local_apply.go:81
tern.(*LocalClient).findBlockingTask     .../pkg/tern/local_apply.go:55
tern.(*LocalClient).checkActiveTaskConflict .../pkg/tern/local_apply.go:26
tern.(*LocalClient).Apply                .../pkg/tern/local_client.go:1725
```

`tryResolveStaleTask` logged `result.State` / `result.Message` in a debug line placed **before** the `if err != nil` check:

```go
result, err := eng.Progress(ctx, &engine.ProgressRequest{...})
c.logger.Debug("conflict check: engine progress", ..., "engine_state", result.State, "message", result.Message, "err", err) // ← line 81
if err != nil {
    return false
}
```

`Engine.Progress` returns a **nil** result alongside a non-nil error on many paths (DSN parse failure, current-schema fetch failure, a DB connection torn down during shutdown, etc.). On any such error, the debug log dereferenced the nil `result` and panicked, crashing the Apply RPC.

## Fix

Move the debug log below the error check so `result` is only dereferenced when non-nil, and log the error on its own line.

## Test

Adds `TestConflictCheckHandlesProgressError`, which drives `tryResolveStaleTask` with an engine that returns `(nil, err)` and asserts it does not panic, returns `false` (task stays blocking), and leaves the task state untouched. Verified the test panics on `main` without the fix and passes with it.

### DIFF
--- a/pkg/tern/local_apply.go
+++ b/pkg/tern/local_apply.go
@@ -78,10 +78,12 @@ func (c *LocalClient) tryResolveStaleTask(ctx context.Context, t *storage.Task, 
 		Database:    database,
 		Credentials: c.credentials(),
 	})
-	c.logger.Debug("conflict check: engine progress", "task_id", t.TaskIdentifier, "engine_state", result.State, "message", result.Message, "err", err)
 	if err != nil {
+		// result is nil on error, so it must not be dereferenced here.
+		c.logger.Debug("conflict check: engine progress failed", "task_id", t.TaskIdentifier, "err", err)
 		return false
 	}
+	c.logger.Debug("conflict check: engine progress", "task_id", t.TaskIdentifier, "engine_state", result.State, "message", result.Message)
 
 	// Engine says terminal — update storage and unblock.
 	// IMPORTANT: Only trust terminal states, NOT "No active schema change".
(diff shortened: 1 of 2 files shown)


aparajon (approved):

pkg/tern/local_apply.go
```diff
@@ -78,10 +78,12 @@ func (c *LocalClient) tryResolveStaleTask(ctx context.Context, t *storage.Task,
 		Database:    database,
 		Credentials: c.credentials(),
 	})
-	c.logger.Debug("conflict check: engine progress", "task_id", t.TaskIdentifier, "engine_state", result.State, "message", result.Message, "err", err)
 	if err != nil {
+		// result is nil on error, so it must not be dereferenced here.
+		c.logger.Debug("conflict check: engine progress failed", "task_id", t.TaskIdentifier, "err", err)
```
aparajon: Should we raise the log level?

morgo:
🤖 Addressed in a34a4c1:
- Reworded both comments to say the Progress result *may* be nil on error — the invariant is not dereferencing `result` when `err != nil`. (Copilot)
- Raised the progress-error log from `Debug` to `Warn` — an engine `Progress` failure during the conflict check is unexpected and worth surfacing. (@aparajon)

The Copilot terminology note about "runningMigration" on line 91 is pre-existing context that this PR doesn't touch, so I left it out of scope.

PR #595, opened 2026-06-29
fix(config): accept deprecated driver keys to prevent crashloop

Kiran01bm:
## Summary
The server config decoder runs with `KnownFields(true)`, so any unknown YAML key is a hard parse error before validation runs. A prior cleanup removed the `operator_workers` / `scheduler_workers` aliases, but deployed configs still set them — so on rollout the config fails to load and the server crashloops.

## What
Re-accept both deprecated keys: restore the struct fields (so the strict decoder no longer rejects them) and the fold logic in `Validate()`, which maps the legacy value into `drivers`, logs a deprecation warning, and rejects setting more than one of the three keys.

## Why
A config knob removal must not be able to crashloop a running fleet whose configs haven't been migrated yet. Folding (rather than ignoring) the legacy key preserves the operator's intended concurrency value. `KnownFields(true)` stays strict for genuine typos. The aliases can be removed again once deployed configs no longer set them.

## Before / after

```text
config: scheduler_workers: 4
  BEFORE                              AFTER
  KnownFields(true) decode           field is known → decode OK
    └─ ✗ "parse config file"           └─ Validate(): fold → drivers=4
         └─ 💥 crashloop on rollout         + warn "deprecated; use drivers"
```

A regression test loads a real config setting each legacy key end-to-end and asserts it folds to `drivers=4` instead of failing.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -622,7 +622,7 @@ The operator is part of Tern orchestration. The API service starts it on server 
 
 Each operator driver runs immediately on startup, then polls every 10 seconds. The driver claims at most one apply per tick and resumes it through the Tern client for that apply's deployment and environment. Fresh applies also send a best-effort wake signal after their apply and task rows are committed, so a driver can claim them immediately instead of waiting for the next poll tick.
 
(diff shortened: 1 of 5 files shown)


Kiran01bm:
closing as per our internal conversation

PR #664, opened 2026-07-02
fix(github): acknowledge a command only when a deployment acts on it

aparajon:
## Why this matters

The 👀 acknowledgment fired at dispatch, before ownership was known. On a multi-tenant repo an unscoped command fans out, so every deployment that *heard* the command reacted — a single comment collected one reaction per installed deployment, reading as "N things are applying" when only one was.

## What it does

Acknowledgment timing follows what is decidable when:

```
command comment
   ├─ repo has NO aggregate role            → 👀 at dispatch (sole deployment,
   │                                          nothing to disambiguate)
   ├─ command is -t-scoped                  → 👀 at dispatch by the addressed
   │                                          tenant (the tenant gate already
   │                                          filtered everyone else)
   └─ unscoped on an aggregate-role repo    → ownership is unknowable until
         │                                    the handler's act-point:
         ├─ deployment silently skips       → skip log only, no reaction
         └─ deployment commits to acting    → 👀 (plan/apply: discovery
            resolves to owned schema · apply-confirm: lock ownership ·
            rollback: apply found + environment-routed · rollback-confirm:
            pinned rollback lock · unlock: locks matched · lifecycle
            controls: apply loaded)
```

Two deliberate consequences of "acknowledgment means acting": the dispatch acknowledgment fires after the flag-rejection checks, so a command rejected for an unsupported flag gets its error reply without a reaction; and handlers that respond with a rejection before their act-point (a lock conflict, a rollback-owned lock, no locks found) likewise reply without acknowledging — the reply itself is the feedback.

For plan and apply the act-point is config discovery plus the in-memory registry lookup — ownership is decidable before any schema files load, so even databases with large schema directories acknowledge within a couple of GitHub API calls of the command. Standard repos and every scoped command keep instant feedback (rollback and lifecycle controls require `-t` in tenant mode, so the whole scoped workflow stays instant); only unscoped discovery-bound commands on shared repos wait for ownership, and there the reaction count finally reflects the deployments doing work. Tests pin the behaviors: dispatch ack on a role-less repo (scoped and bare multi-env plan), dispatch ack for a scoped command on a participant, act-point ack for an owning deployment, and no ack for a silently-skipping one. The bare `schemabot plan` follows the same contract as its scoped siblings — dispatch ack where ownership is unambiguous, discovery act-point ack on aggregate-role repos.

### DIFF
--- a/pkg/webhook/aggregate_fanout_silent_skip_test.go
+++ b/pkg/webhook/aggregate_fanout_silent_skip_test.go
@@ -309,6 +309,24 @@ func TestCommandAcknowledgmentFollowsOwnership(t *testing.T) {
 		}
 	})
 
+	// A bare multi-env plan on a repo without a schema-dir allowlist resolves
+	// config discovery successfully even on a deployment that does not own the
+	// database — ownership is only decided at the registry lookup. A fan-out
+	// participant in that position silently skips, so it must not acknowledge.
+	t.Run("multi-env plan on unowned database does not react", func(t *testing.T) {
+		h, mux, _ := newFanOutSkipHandler(t, aggregateLeaderConfig())
+		serveSchemaConfigForDatabase(t, mux, "orders")
+		reactions := registerReactionRecorder(t, mux)
+
+		h.handleMultiEnvPlan("octocat/hello-world", 1, "", "", 12345, "hubot", false, true, 42)
+
+		select {
+		case <-reactions:
+			t.Fatal("a deployment that cannot resolve the database in its registry must not acknowledge")
+		case <-time.After(100 * time.Millisecond):
+		}
+	})
+
 	t.Run("silently skipping deployment does not react", func(t *testing.T) {
 		h, mux, comments := newFanOutSkipHandler(t, aggregateLeaderConfig())
 		serveSchemaConfigForDatabase(t, mux, "orders")
--- a/pkg/webhook/aggregate_fanout_silent_skip_test.go
+++ b/pkg/webhook/aggregate_fanout_silent_skip_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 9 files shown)


Kiran01bm:
**Verdict: one real (narrow) correctness gap; everything else checks out**
The finding: handleMultiEnvPlan acknowledges before ownership is fully
  decided (pkg/webhook/plan.go:207)
  
**Minor notes (non-blocking):**
  1. acknowledgeCommandActPoint's nil-service guard (h.service != nil && role
  == "") doesn't return early when h.service == nil, while the dispatch
  condition acks in that case — both would fire. GitHub's add-reaction API is
  idempotent per user+content so the effect is invisible, but the predicate
  should mirror dispatch exactly.
  2. The probe duplicates discovery API calls (FetchPullRequest, FetchPRFiles,
  config fetches) on every unscoped plan/apply, per deployment, on aggregate
  repos. Accepted trade-off per the description; worth keeping in mind for
  rate limits on busy shared repos.
  3. Behavior change worth flagging in the description: the dispatch ack now
  fires after the flag-rejection checks (-y, --defer-cutover, unsupported -d),
  so those error replies no longer get 👀 even on single-deployment repos —
  the old ack preceded them. Likewise, handlers that "act" by posting a
  rejection before their act-point (lock conflict, rollback-owned lock,
  no-locks-found) respond without acknowledging. Both look intentional under
  "ack = doing the work," just making the change visible.

Kiran01bm (approved):

aparajon:
🤖 Thanks — the main finding was exactly right, and subtle: `configPathManagedByRepo` short-circuits on repos without a schema-dir allowlist, so config discovery alone could pass on a database another deployment owns, and the multi-env ack fired before the registry lookup decided real ownership. Fixed in f22b5772 — the acknowledgment now follows the registry/environment resolution (in-memory lookups, so it's still immediate), with a test pinning that a deployment that can't resolve the database in its registry never reacts.

On the minors: (1) the act-point nil-service guard now mirrors the dispatch predicate exactly (same commit), so the two can never both fire even in harness setups; (2) agreed on the duplicated discovery calls — accepted trade-off, and worth revisiting if the rate-limit gauges ever show pressure on busy shared repos; (3) both behavior changes are now called out in the description — flag-rejection replies and pre-act-point rejections respond without a reaction, deliberately, since the reply itself is the feedback.

One thing your finding surfaced in passing: the multi-env error branches post their rejection comments without the unowned-silent-skip gate the single-env call sites have — that's pre-existing on main, out of scope here, and tracked separately.

— Claude (claude-fable-5) via Claude Code, with maintainer approval

PR #689, opened 2026-07-08
fix(github): fall back to configured schema dirs when the repo tree is truncated

aparajon:
## Why this matters

Repositories beyond GitHub's Trees API cap get a truncated recursive listing, so every whole-repo config scan (bare `schemabot plan`, `plan -e`, `plan -d` for a database outside the diff, unlock inference) fails closed with an error the user cannot act on. In the largest monorepos, that turns routine self-service commands into operator escalations.

## How it works

SchemaBot's server config already says where configs are *allowed* to live (`databases.<db>.allowed_dirs`) — a config anywhere else would be rejected by policy. So when the whole-repo listing truncates, `FindAllConfigs` doesn't need it: it walks directly to each configured directory and reads just that. The schema-file loader recovers the same way.

Git makes this possible because every directory is its own tree object listing only its *immediate* children, with pointers to its subdirectories' trees. A one-level fetch is only as big as the directory itself, so it never hits the cap — only the flattened whole-repo listing does. Resolving `apps/widgets/schema` is just following pointers, like `cd apps && cd widgets && cd schema && ls -R` instead of `ls -R` from the repo root:

```
repo root — 49k files; recursive listing → TRUNCATED ✗

fetch root, 1 level (~40 entries)     ── 1 small call
  └─ "apps" → tree a1b2c3
fetch apps, 1 level                   ── 1 small call
  └─ "widgets" → tree d4e5f6
fetch widgets, 1 level                ── 1 small call
  └─ "schema" → tree 789abc
fetch schema, RECURSIVE (rooted here) ── only this directory
  └─ schemabot.yaml, users.sql, …        must fit under the cap ✓
```

Shallow per-level fetches are memoized per scan, so hint directories sharing a path prefix (e.g. `apps/widgets/schema` and `apps/payments/schema`) reuse the root and `apps` levels instead of re-walking them.

The fallback only runs when the configured directories are **provably exhaustive** — every location a policy-valid config could live in is covered. If any repo-eligible database has no `allowed_dirs`, or a wildcard (`*`) or repo-root (`.`) entry, its config could legally live anywhere, and a partial probe returned as authoritative could plan the wrong database (one config found where a full scan would find several) or report a misleading "database not found". In those cases — and when no directories are configured, none contain a config, or the schema path is the repo root — discovery keeps failing closed with the truncation error. Symlink handling also still fails closed.

```
FindAllConfigs(repo, ref)
  ├─ full recursive tree ── ok ───────────────→ scan for schemabot.yaml
  └─ truncated
       ├─ hints exhaustive (every db has literal allowed_dirs)
       │    └─ per-dir subtree walk ── found ─→ configs
       └─ no hints / not exhaustive / nothing found
                                     ─────────→ fail closed (ErrGitTreeTruncated)
```

Schema changes stay self-service at any repository size: the fail-closed guarantee is unchanged, but a truncated tree no longer strands users whose schema directories are perfectly listable on their own.

---

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -737,6 +737,40 @@ func (c *ServerConfig) RepoAdmins(repo string) (teams, users []string) {
 	return repoConfig.AdminTeams, repoConfig.AdminUsers
 }
 
+// SchemaDirHintsForRepo returns the configured schema directories of every
+// database that accepts changes from repo, sorted and deduplicated. Config
+// discovery uses these as probe targets when GitHub truncates the repository
+// tree (repos beyond the Trees API response cap), where a whole-repo scan is
+// impossible. Wildcard and empty entries carry no directory to probe and are
+// skipped; a database with no allowed_repos restriction accepts any trusted
+// repo, so its directories are always included.
+func (c *ServerConfig) SchemaDirHintsForRepo(repo string) []string {
+	seen := make(map[string]struct{})
+	var dirs []string
+	for _, db := range c.Databases {
+		if len(db.AllowedRepos) > 0 && !repoAllowed(db.AllowedRepos, repo) {
+			continue
+		}
+		for _, dir := range db.AllowedDirs {
+			// Normalize exactly like the allowed_dirs policy checks so the
+			// probed directories match what the policy would accept. Wildcard
+			// and repo-root entries name no directory to probe; invalid
+			// entries are rejected at config load and cannot appear here.
+			normalized, err := normalizeSchemaPath(dir)
+			if err != nil || normalized == "*" || normalized == "." {
+				continue
+			}
+			if _, ok := seen[normalized]; ok {
+				continue
+			}
+			seen[normalized] = struct{}{}
+			dirs = append(dirs, normalized)
+		}
+	}
+	slices.Sort(dirs)
+	return dirs
+}
+
 // DeploymentTarget is one entry in EnvironmentConfig.Deployments. It carries
 // the per-deployment override values for a multi-deployment environment. The
 // enclosing map key identifies the Tern deployment (and must also appear in
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -741,23 +741,39 @@ func (c *ServerConfig) RepoAdmins(repo string) (teams, users []string) {
 // database that accepts changes from repo, sorted and deduplicated. Config
 // discovery uses these as probe targets when GitHub truncates the repository
 // tree (repos beyond the Trees API response cap), where a whole-repo scan is
-// impossible. Wildcard and empty entries carry no directory to probe and are
-// skipped; a database with no allowed_repos restriction accepts any trusted
-// repo, so its directories are always included.
-func (c *ServerConfig) SchemaDirHintsForRepo(repo string) []string {
+// impossible. A database with no allowed_repos restriction accepts any
+// trusted repo, so its directories are always included.
+//
+// exhaustive reports whether the returned directories cover every location a
+// policy-valid config could live in. A database with no allowed_dirs
+// restriction, or with a wildcard ("*") or repo-root (".") entry, accepts a
+// config in directories no hint list can cover — for such a repo the
+// truncated-tree fallback must keep failing closed, because a partial probe
+// could return a confidently wrong result (a single config where a full scan
+// would find several, or a "database not found" for a config that exists).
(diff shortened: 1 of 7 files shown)


Kiran01bm:
**Finding 1: the fallback's completeness assumption doesn't hold for every policy shape**

The fallback justifies returning a partial result with (comment in findConfigsInHintDirs): "a config outside every configured directory would be rejected by the allowed_dirs policy anyway." That's only true for databases with a non-empty, literal allowed_dirs.
Three configuration shapes break it:

Empty allowed_dirs — source_policy.go:103 enforces the restriction only if len(dbConfig.AllowedDirs) > 0; an empty list means any directory is accepted. Such a database contributes zero hints but its config anywhere in the repo is policy-valid.
"*" entry — schemaPathAllowed treats it as allow-all; SchemaDirHintsForRepo (correctly) skips it as a probe target, with the same effect.
"." entry — policy accepts a repo-root config, but hints skip ., so a root config is invisible to the fallback.
The failure mode only appears when the fallback partially succeeds — one hinted dir yields a config, so ok=true and the incomplete result is returned as authoritative. Then:

FindConfigInRepo (bare schemabot plan single-config inference, pkg/github/config.go:482) can report exactly one config where a full scan would have returned ErrMultipleConfigs — the wrong database gets planned silently. That's confident-wrong replacing
fail-closed.
FindConfigByDatabaseName (plan -d) returns DatabaseNotFoundError with an "available databases" list that omits a database whose config genuinely exists — a misleading error replacing the actionable truncation error.
Suggested fix: make the code comment true by construction — gate the fallback on the hints being provably exhaustive for the repo. E.g., have SchemaDirHintsForRepo return (dirs, exhaustive bool) where exhaustive is false if any repo-eligible database has empty
allowed_dirs or a */. entry, and keep failing closed when not exhaustive. If every production database in large repos already sets literal allowed_dirs, the gap is latent today — but nothing validates that, and the PR's "fail-closed guarantee is unchanged"
claim quietly depends on it.

**Finding 2 (minor): repo-root schema path silently returns no files**

In fetchSchemaFilesFromTree, a config at the repo root has schemaPath == "."; on a truncated repo, FetchGitSubtree(…, ".") walks segments looking for a tree entry literally named ., finds none, and returns found=false → the function returns nil, nil ("schema
path does not exist") instead of the pre-PR truncation error. Downstream that reads as "no schema files" rather than "repo too large". Cheap guard: treat "."/empty schemaPath as unresolvable in the truncated branch and keep the truncation error.

**Verified sound**

Fail-closed is preserved when the fallback finds nothing (ok=false → ErrGitTreeTruncated), and errors.Is(…, ErrGitTreeTruncated) still matches through the new wrapping. Invalid-config-only results correctly count as "found" so users get the parse error, not
truncation.
Subtree fetches are double-guarded (shallow-level truncation and recursive-subtree truncation both fail closed — covered by tests), and the symlink fail-closed path survives the fallback since the existing mode-120000 filter runs on the prefixed entries.
Dedup across nested hint dirs works — both scans produce the identical repo-relative configPath string. isConfigFile matches on path.Base, so relative subtree paths are fine. Hint normalization (normalizeSchemaPath) is exactly the policy-side normalization,
so probed dirs match what policy would accept; allowed_dirs has no glob support beyond bare *, so no literal-glob probing surprises.
Wiring covers both single-app and multi-app serve paths; hints are sorted/deduped so probe order is deterministic.

**One non-correctness note:**
The per-dir segment walk re-fetches shared prefix levels (the root tree is fetched once per hint dir) — for many databases with deep common prefixes that's redundant API volume against a rate limit; a per-(ref, path) memo inside one
FindAllConfigs call would fix it.

Kiran01bm (approved):

aparajon:
🤖 All three addressed in 9891ec10. Finding 1: `SchemaDirHintsForRepo` now returns `(dirs, exhaustive)` exactly as you suggested — exhaustive is false when any repo-eligible database has no `allowed_dirs` or a `*`/`.` entry, and `findConfigsInHintDirs` refuses to probe non-exhaustive hints, so the "would be rejected by policy anyway" comment is now true by construction. Finding 2: a repo-root schema path keeps the truncation error in the truncated branch instead of misreporting an empty directory. The rate-limit note too: the hint scan memoizes shallow per-level fetches by tree SHA, so shared prefixes cost one call per level per scan. Tests cover each policy shape. — Claude (claude-fable-5) via Claude Code

PR #701, opened 2026-07-10
fix(github): fast-ack pr auto-plan webhooks

Kiran01bm:
## Summary
Pull request webhooks now acknowledge quickly after validation/routing and dispatch auto-plan bootstrap, discovery, and check publication asynchronously. This keeps SchemaBot responses away from GitHub's fixed 10-second webhook delivery timeout, which is not configurable by the receiving service.

This is phase one: it reduces the synchronous webhook response path and adds delivery-GUID correlation plus a bootstrap-failure metric for async completion/failure visibility. Durable enqueue/dedup, bounded workers, graceful shutdown draining, redundant convergence signals, and check backfill remain follow-up safety nets for ack-then-crash, webhook-storm, or lost-delivery cases.

## What
- Move `pull_request` auto-plan bootstrap/discovery/check publication behind the existing safe async webhook runner.
- Return `auto-plan started` quickly after the webhook is accepted.
- Include the GitHub delivery GUID in auto-plan triggered/dispatched/error logs for incident correlation.
- Emit an `auto_plan_bootstrap_failed` webhook metric when fast-acked async work cannot bootstrap.
- Add repo/PR/install identifiers to async panic logs.
- Harden integration tests so they assert async outcomes instead of only the immediate 200 response.

## Why
GitHub webhook deliveries have a hard 10-second response timeout, and that timeout is not configurable. Auto-plan work can include GitHub client setup, PR file reads, config discovery, stale-check cleanup, and aggregate check publishing, so doing that work before returning risks GitHub recording a timed-out delivery.

This PR is the tactical fast-ack mitigation; it is not the final durable ingestion design.

```text
Before:
GitHub pull_request webhook
        │
        ▼
SchemaBot validates + bootstraps + discovers configs + refreshes checks
        │
        ▼
HTTP 200 only after synchronous work completes
```

```text
After:
GitHub pull_request webhook
        │
        ▼
SchemaBot validates/routes + starts async auto-plan
        │
        ├──▶ HTTP 200: auto-plan started
        │
        ▼
Async bootstrap + discovery + check/comment publication
        │
        ▼
Logs dispatch/failure with X-GitHub-Delivery correlation
```

## Follow-ups
- Add a tighter deadline or earlier handoff around any remaining pre-dispatch work that can consume webhook response budget.
- Add graceful shutdown draining for in-process webhook goroutines, or replace them with durable workers.
- Persist durable webhook jobs before acknowledgement, with dedup, replay, attempts, and last-error state.
- Run auto-plan/check publication through bounded workers or per-repo/PR single-flight to avoid webhook storms exhausting GitHub/API budgets.
- Add redundant convergence signals into the same durable work-item path, keeping work PR-scoped.
- Treat checks backfill as the safety net for `200 ack but missing checks`, since failed-delivery redrive only helps when GitHub records a failed delivery.
- Add started/dispatched/failed metrics keyed by delivery GUID and alert on missing completion or missing expected checks.

## References
- GitHub webhook timeout docs: https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/troubleshooting-webhooks#timed-out

### DIFF
--- a/pkg/webhook/aggregate_check_integration_test.go
+++ b/pkg/webhook/aggregate_check_integration_test.go
@@ -12,6 +12,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -637,12 +638,9 @@ func TestE2EParticipantSilentOnNonSchemaPR(t *testing.T) {
 
 	// PR changed files — no schema files.
 	filesFetched := make(chan struct{})
+	var filesFetchedOnce sync.Once
 	mux.HandleFunc("GET /repos/octocat/hello-world/pulls/1/files", func(w http.ResponseWriter, _ *http.Request) {
-		select {
-		case <-filesFetched:
-		default:
-			close(filesFetched)
-		}
+		filesFetchedOnce.Do(func() { close(filesFetched) })
 		_ = json.NewEncoder(w).Encode([]*gh.CommitFile{
 			{Filename: new("README.md"), Status: new("modified")},
 		})
--- a/pkg/webhook/aggregate_check_integration_test.go
+++ b/pkg/webhook/aggregate_check_integration_test.go
@@ -590,7 +590,7 @@ func TestE2EPassingAggregateOnNonSchemaPR(t *testing.T) {
 	h.ServeHTTP(rr, req)
 
 	require.Equal(t, http.StatusOK, rr.Code)
-	assert.Contains(t, rr.Body.String(), "no schema files in PR")
+	assert.Contains(t, rr.Body.String(), "auto-plan started")
 
 	// Wait for both passing aggregates (staging + production)
 	seen := map[string]bool{}
@@ -636,7 +636,13 @@ func TestE2EParticipantSilentOnNonSchemaPR(t *testing.T) {
 	})
 
 	// PR changed files — no schema files.
+	filesFetched := make(chan struct{})
 	mux.HandleFunc("GET /repos/octocat/hello-world/pulls/1/files", func(w http.ResponseWriter, _ *http.Request) {
+		select {
+		case <-filesFetched:
+		default:
+			close(filesFetched)
+		}
 		_ = json.NewEncoder(w).Encode([]*gh.CommitFile{
 			{Filename: new("README.md"), Status: new("modified")},
 		})
@@ -672,10 +678,16 @@ func TestE2EParticipantSilentOnNonSchemaPR(t *testing.T) {
 	h.ServeHTTP(rr, req)
 
 	require.Equal(t, http.StatusOK, rr.Code)
-	assert.Contains(t, rr.Body.String(), "aggregate participant, staying silent")
+	assert.Contains(t, rr.Body.String(), "auto-plan started")
 
-	// The skip is synchronous (it returns before scheduling any passing
-	// aggregate), so a brief drain confirms no check run was posted.
+	// Wait until async config discovery reads the PR files, then drain
+	// briefly to confirm the participant stayed silent instead of publishing an
+	// aggregate on a PR with no managed schema changes.
+	select {
+	case <-filesFetched:
+	case <-time.After(webhookIntegrationCheckRunDeadline):
+		t.Fatal("timed out waiting for participant auto-plan discovery")
+	}
 	select {
 	case cr := <-checkRuns:
 		t.Fatalf("participant posted a check run on a non-schema PR: %q", cr.Name)
@@ -1151,7 +1163,7 @@ func TestE2EPassingAggregateOnSQLWithoutSchemabotYAML(t *testing.T) {
 	h.ServeHTTP(rr, req)
 
 	require.Equal(t, http.StatusOK, rr.Code)
-	assert.Contains(t, rr.Body.String(), "no schema files in PR")
+	assert.Contains(t, rr.Body.String(), "auto-plan started")
 
 	// Should post passing aggregate even though .sql files changed
 	select {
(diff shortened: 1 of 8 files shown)


aparajon:
🤖 Adversarial correctness review (built the head, ran the six touched integration tests with `-race` against real MySQL containers — all pass; audited against the AGENTS.md self-review bar).

**What checks out.** The context plumbing is right: the async closure never touches the request context — `autoPlanBootstrap` builds a fresh `context.Background()` + 30s budget, unchanged from before, so nothing inherits the dead request deadline. Everything that must stay synchronous did: HMAC verification, the App-owns-repo cross-check, the repo allowlist, and repo-level installation resolution all still run before the ack and still fail closed with 4xx/5xx — those failures remain visible to GitHub and redriveable. Fail-closed check behavior is preserved and the tests prove it: GitHub-unavailable during discovery still publishes failing aggregates with the machine-readable blocking reason, and the unverified-head case still publishes nothing. No test functions were deleted — every replaced sync-response assertion was *upgraded* (stored-check `Eventually`, sync-point channels on the mock endpoints, negative drains anchored on the files fetch). Per-PR concurrency is not new (GitHub already delivers concurrently across connections and pods), so fast-ack doesn't create the stale-SHA race; existing guards govern it. `TestE2EAutoPlanBootstrapDoesNotBlockWebhookAck` is a genuinely good test — it proves the ack can't wait on GitHub client construction.

**Findings:**

1. **Failure visibility regresses from "GitHub-recorded failed delivery" to "one error log."** Before this change, a bootstrap failure returned 500 → GitHub records a failed delivery → visible in the App's delivery UI *and* recoverable via `schemabot webhooks redrive`. Now it's a 200 + `logger.Error`, invisible to both. `schemabot checks backfill` (#692, now merged) is the right recovery tool for the "200 ack but missing checks" class — but recovery still needs a *trigger*. Two acceptable shapes: (a) a metric counter on the async bootstrap-failure branch (per the "metrics should be actionable" bar for rare dangerous branches), or (b) a documented log monitor on the invariant this PR already creates — the `auto-plan triggered` / `auto-plan finished` pair is keyed by `delivery_id`, so *triggered lines with no matching finished line* detect bootstrap failures, hangs, **and** the shutdown-dropped case from finding 2, which emits no error log at all. Option (b) only works end to end if the log gaps in findings 3–4 are closed.

2. **No graceful-shutdown drain — routine deploys now drop acked auto-plans.** There's no WaitGroup in the handler; `http.Server.Shutdown` drains in-flight *requests*, and this work no longer lives in one. Before, discovery + the fail-closed check decisions completed inside the request and were protected by the drain; now a SIGTERM mid-roll silently kills them after the 200. Rolling deploys are routine, so this is the most probable real-world loss path — PRs opened during a roll can sit checkless until a push/rerun/backfill. Merge safety holds (a missing required aggregate blocks merge), but the fix is cheap: track `goSafe` goroutines in a WaitGroup and drain on shutdown, bounded by the existing 30s ctx.

3. **`recoverPanic` gaps: no identifiers in the log, raw panic in the comment.** The helper receives `repo`, `pr`, and `installationID` but only uses them for the PR comment — the `goroutine panic` log line is just error + stack, so a panic in the new async path can't be correlated to a PR or delivery without parsing the stack trace, and it leaves a triggered-without-finished pair with no explanation nearby. Adding `repo`/`pr`/`installation_id` to the log attrs fixes every `goSafe` call site at once. Separately (pre-existing, fine as a follow-up): it posts the raw panic value into a public PR comment; per the untrusted-error-strings rule the comment should be a fixed sanitized line with the detail logged server-side.

4. **`delivery_id` stops at the wrapper, and "finished" means "dispatched."** The error logs inside `runAutoPlanForPR` (`failed to fetch PR files for auto-plan`, `failed to discover configs for PR`) don't carry the delivery GUID — correlating them means joining on repo/pr/head_sha, which is ambiguous when two deliveries for the same PR are in flight (rapid synchronize events). Threading `deliveryID` in (or a derived logger) closes it. Also, on the configs-found path `auto-plan finished` fires when *dispatch* completes — the per-database work continues in further `goSafe` goroutines with no delivery-scoped completion log (pre-existing shape, but this PR makes the finished line the triage anchor). Cheapest fix is naming honesty: `auto-plan dispatched`, or an explicit message that per-database planning continues async. Related operator-facing trade: GitHub's delivery UI now shows "auto-plan started" for every outcome; the real decision only lands in this log.

**Verdict: needs-changes** — the async mechanics are sound and well-tested; what's left is a detection story for finding 1 (metric or monitored log invariant, with 3–4's identifier gaps closed) and a decision on the shutdown drain (2), either here or as an explicitly tracked fast-follow given the "phase one" framing. Both Copilot comments are already addressed on the head.

*Edited: (1) originally said #692 was an open draft with a merge-order dependency — it merged 2026-07-09; reworded finding 1 accordingly. (2) expanded findings 1, 3, 4 with the logs-as-detection angle: the triggered/finished `delivery_id` pair as a monitorable invariant, plus the identifier gaps that keep it from working end to end.*

aparajon (approved):

Kiran01bm:
> 🤖 Adversarial correctness review (built the head, ran the six touched integration tests with `-race` against real MySQL containers — all pass; audited against the AGENTS.md self-review bar).

| # | Review comment | Response / disposition |
|---:|---|---|
| 1 | Bootstrap failure visibility regresses from GitHub-recorded failed delivery to only an error log. | Agreed. I added an `auto_plan_bootstrap_failed` webhook metric on the async bootstrap-failure path, so this class has an alertable signal even though GitHub records the delivery as `200`. |
| 2 | No graceful shutdown drain; routine deploys can drop already-acked auto-plan goroutines. | Agreed. I’m treating this as a follow-up rather than expanding this tactical PR: `goSafe` is a generic async webhook helper used by multiple paths, and the cleaner fix is either a lifecycle-aware drain across those goroutines or replacing them with durable workers. |
| 3 | `recoverPanic` posts raw panic values into public PR comments. | Agreed. This is pre-existing, but fast-ack routes more work through the async recovery path. I’m treating this as a follow-up to keep raw panic values/stack traces in server logs and post only a fixed user-safe message publicly. |
| 4 | GitHub delivery UI now shows `auto-plan started` for every accepted webhook; final outcome is only in logs. | Agreed; this is the expected fast-ack tradeoff. The PR keeps `X-GitHub-Delivery` on triggered/finished/error logs, and the bootstrap-failure path now emits a metric for the failure mode that would otherwise be easiest to miss. |

PR #758, opened 2026-07-18
feat(github): decay progress-comment edit cadence as the comment ages

aparajon:
## Why this matters

An actively copying apply edits its PR progress comment every 5 seconds — ~720 writes per hour, per apply, for as long as the copy runs. Nobody watches a PR comment continuously for a day, and a single week-long schema change can, by itself, keep the installation near GitHub's secondary content-creation limit. The CLI is the high-resolution progress surface; the PR comment needs to stay current, not real-time.

## What it does

Spaces progress-comment edits further apart as the live progress comment ages:

| comment age | edit interval |
|---|---|
| < 10 minutes | 5s |
| < 1 hour | 30s |
| < 2 hours | 1m |
| < 4 hours | 2m |
| < 6 hours | 5m |
| older | 10m |

- The clock is anchored to the **live progress comment**, not the apply: a rotation (resume after a stop, volume change) posts a fresh comment because someone just acted on the apply, so the fresh comment starts back at the fastest cadence. A recovered drive also restarts fast.
- State transitions bypass the cadence entirely — a slowdown never delays "Waiting for Cutover" or a terminal state reaching the PR.
- The comment states its own cadence: the "Last updated" footer renders `· updates every ~2m` for the current rung, so a reader watching a long apply sees the widening gap as designed behavior rather than a stall. The note re-renders once at each rung transition, keeping the advertised cadence truthful.

Stacked on #757.

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -2145,7 +2145,7 @@ To stop this schema change:
 schemabot stop apply-a1b2c3d4e5f6 -e staging
 ```
 
-_Last updated: <relative-time datetime="2026-01-01T00:00:00Z">2026-01-01 00:00:00 UTC</relative-time> (2026-01-01 00:00:00 UTC)_
+_Last updated: <relative-time datetime="2026-01-01T00:00:00Z">2026-01-01 00:00:00 UTC</relative-time> (2026-01-01 00:00:00 UTC) · updates every ~5s_
 
 </details>
 
--- a/pkg/webhook/apply.go
+++ b/pkg/webhook/apply.go
@@ -189,8 +189,13 @@ func shardCommentTableKey(applyOperationID *int64, namespace, table string) stri
 // formatProgressComment renders the progress comment using the template system.
 // It is the no-operations fallback (load error, or the initial rollback comment),
 // so it carries no VSchema — the observer refreshes VSchema once operations load.
-func formatProgressComment(apply *storage.Apply, tasks []*storage.Task, shardsByTable map[string][]*storage.Task, tenant string) string {
-	return templates.RenderApplyStatusComment(buildApplyCommentData(apply, tasks, operationDisplay{}, shardsByTable, tenant))
+// updateCadence is the current spacing between automatic refreshes of the
+// comment, rendered beside the "Last updated" footer; zero renders no cadence
+// note (terminal finalizations).
+func formatProgressComment(apply *storage.Apply, tasks []*storage.Task, shardsByTable map[string][]*storage.Task, tenant string, updateCadence time.Duration) string {
+	data := buildApplyCommentData(apply, tasks, operationDisplay{}, shardsByTable, tenant)
+	data.UpdateCadence = updateCadence
+	return templates.RenderApplyStatusComment(data)
 }
 
 // formatSummaryComment renders the final summary comment for a terminal apply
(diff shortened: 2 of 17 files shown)


morgo (approved):
🤖 Approving on Morgan's behalf (automated review).

The motivation is sound and the numbers back it: ~720 edits/hour/apply against a shared per-installation write budget, for a surface nobody watches continuously. Comment rendering only, no MySQL path, full suite green.

I verified the two safety claims against source rather than the table:

**State transitions genuinely bypass the cadence.** The stagnant branch is only reachable when `totalRows == o.lastRowsCopied && currentState == o.lastState`, so any state change falls to the else branch, where the early return is guarded by `&& currentState == o.lastState` and therefore doesn't fire. A slowdown can't delay "Waiting for Cutover" reaching the PR. Terminal is separate again — `OnTerminal` renders with cadence `0` and edits directly, never through `OnProgress`.

**The advertised cadence can't go stale silently.** This depends on an interaction with #757 that isn't obvious: `CommentContentKey` only masks the rendered timestamp, so the ` · updates every ~2m` note is part of the content key. A rung transition therefore changes the key and the edit survives the unchanged-content dedup instead of being skipped. Had the key normalized the footer, the comment would have kept advertising a cadence it no longer honored. It works — worth knowing it works *because* of how the key is built.

Also correct: the cutover comment renders with cadence `0` (it's frozen pending an operator, so advertising a refresh rate would be a lie), and both rotation paths reset `progressCommentSince` while rendering with `freshProgressCadence()`, so a fresh comment's advertised rate matches the rung it actually restarts on.

**One finding: the decay anchor is in-memory, so it resets on every observer construction.**

```go
progressCommentSince time.Time  // zero until the first progress tick
```

`progressCommentSince` lives only on the `CommentObserver` struct. A new observer starts zero and anchors to its own first tick, so every drive recovery, lease handoff, or pod restart puts the comment back on the 5s rung regardless of how old the comment actually is.

The PR frames this as intentional — "A recovered drive also restarts fast" — and for a *rotation* that's clearly right, since someone just acted on the apply. For a recovery it's a different claim, and it works against the PR's own motivation: the write-budget argument is about aggregate volume over long applies, and the week-long copy cited as the motivating case is precisely the one most likely to span many observer lifetimes. If drives change hands every few minutes on a busy server, the schedule never leaves its first rung and the decay delivers little of what it promises. Nothing about GitHub's rate limit cares how long a particular observer object lived.

The fix is available because the comment itself is durable: the tracked progress comment carries a persisted `GitHubCommentID`, so its real age can come from the stored tracked-comment row (or from GitHub's own `created_at`) rather than from process-local state. Rotation would still reset the anchor by posting a new comment — which is the behavior you want — while recovery would resume at the rung the comment's true age earns.

**Nits, none blocking:**

`freshProgressCadence()` ignores its receiver and returns a constant; as a method it reads as though it might vary per observer. `progressEditInterval` is a getter that mutates `progressCommentSince` on first call — documented as being called under `o.mu`, so it's safe, just surprising at the call site. And `progressEditDecayMaxInterval` (10m) is exactly `unchangedCommentGraceWindow` (10m); if that equality is meaningful it deserves a comment, and if it's coincidence the two will drift apart eventually.

Worth noting this head is from 2026-07-20, so it will likely want a rebase before it lands.

PR #759, opened 2026-07-18
feat(github): add configurable floor for progress-comment edit cadence

aparajon:
## Why this matters

The built-in edit cadence is a fixed policy. During a GitHub rate-limit incident — or on a server hosting many concurrent long-running applies — operators need one valve that slows every apply's progress-comment traffic at once, without a code change or redeploy of new logic.

## What it does

Adds a server config setting that floors the built-in age-decayed cadence:

```yaml
progress_comment_min_edit_interval: 2m  # Go duration; default unset (no floor)
```

- No progress comment is edited more often than the configured interval, regardless of the built-in cadence.
- State changes still publish immediately — the floor spaces out progress noise, not state transitions.
- The cadence the progress comment advertises in its "Last updated" footer reflects the floor, so PR readers see the configured interval rather than wondering why updates slowed down.
- Invalid or negative values are rejected at config validation; an invalid value read at runtime logs a warning and falls back to the built-in cadence.
- Documented in `docs/configuration.md`; users needing high-resolution progress use the CLI, which polls the server directly and is unaffected.

Stacked on #758.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,8 +16,10 @@
 - [Pending Drops](#pending-drops)
 - [Storage Schema Changes](#storage-schema-changes)
 - [Support Channel](#support-channel)
+- [Progress Comment Edit Cadence](#progress-comment-edit-cadence)
 - [Repository Allowlist](#repository-allowlist)
 - [PR Checks Gate](#pr-checks-gate)
+- [Base Branch Schema Freshness](#base-branch-schema-freshness)
 - [Review Gate](#review-gate)
 - [Authentication](#authentication)
   - [OIDC (Bearer tokens)](#oidc-bearer-tokens)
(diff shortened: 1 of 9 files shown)


morgo (approved):
🤖 Approving on Morgan's behalf (automated review).

A single operator valve for GitHub write pressure that needs no redeploy is worth having, and this one is correctly bounded. Comment rendering plus config, no MySQL path, full suite green.

Checked against source:

- **The floor only ever slows.** `if o.minEditInterval > interval { interval = o.minEditInterval }` is a true max over the decayed value, so it can't accidentally speed edits up when set below the current rung — the failure mode I went looking for, since a "minimum interval" applied as an assignment rather than a max would silently override the decay downward on the later rungs.
- **It governs fresh comments too**, and `freshProgressCadence()` applies the same max, so a rotation under a configured floor advertises the floor rather than the 5s it would otherwise claim. Advertised and actual stay in agreement, which is the property #758 established and this preserves.
- **State transitions are untouched.** The floor only feeds `decayed`, which is consulted solely on the paths guarded by `currentState == o.lastState`. A configured floor of 10 minutes still lets a terminal state or a cutover transition land immediately — the thing that would make this valve dangerous if it were wrong.
- **Validation is real and wired.** `ProgressCommentMinEditIntervalDuration` rejects both unparseable and negative values, and `Validate()` calls it, so a bad value fails at startup rather than at the first progress tick. Empty string → zero → no floor is a clean default.

The rewrite of `progressEditInterval` to initialize with `progressEditDecayMaxInterval` and `break` on the first matching rung is also a tidier form of #758's loop.

**One nit, not blocking:** the floor has no upper bound — validation rejects negative but accepts `720h`, which would effectively freeze progress comments. That degrades acceptably rather than dangerously, since state transitions and terminal updates still publish, so the worst case is stale intermediate progress on a surface the config explicitly redirects to the CLI. A sanity cap or a startup warning above some threshold would make a fat-fingered value obvious at deploy time instead of by absence.

This also inherits #758's finding — the decay anchor is process-local, so a floor is the only part of the cadence that survives a drive handoff. That arguably makes this PR *more* useful in the interim, not less: it's the one lever that isn't reset by recovery.

Same note as #758: this head is from 2026-07-20 and will likely want a rebase before landing.

PR #762, opened 2026-07-18
fix(tern): retry cutover quietly when the deploy request has not staged its changes

aparajon:
## Why this matters

PlanetScale can report a deploy request as `pending_cutover` a beat before its staged changes are visible to the apply endpoint. When a cutover fires in that window, the API rejects it with "deploy request changes have not been staged" — an expected, self-clearing condition. Today that rejection is logged as a cutover **failure** (with a misleading "may have been deleted" hint), paging operators for an apply that completes normally on the next progress tick — and a queued cutover command hitting the same window is terminally failed, stranding a deferred cutover until a human notices and re-issues it.

## What it does

- Adds `engine.NotReadyError` / `IsNotReady`: a retryable classification for "the backend advertised the gate before it was ready to accept the operation".
- The PlanetScale engine classifies the not-staged rejection as not-ready and drops the deleted-deploy-request hint from its message, since the deploy request demonstrably exists.
- The drive's auto-cutover records its trigger event once per drive and retries not-ready rejections quietly at Info. A rejection that outlives the staging window records a one-time timeline error and escalates to Error logging on every further tick, so a backend that never stages the cutover still pages.
- A durable cutover request (deferred/operator-initiated) rejected as not-ready stays **pending** — the next progress tick retries it instead of terminally failing the command. The timeline records the trigger only for the attempt the backend accepts (or definitively fails), not one per retry.

```
poll tick: pending_cutover → Cutover()
  ├─ ok                 → cutover proceeds
  ├─ NotReady (staging) → request stays pending, Info log, re-fires next tick
  │    └─ persists past staging window → one timeline error + Error log every tick (pages)
  └─ other error        → terminal failure, Error log (pages)
```

### DIFF
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -143,6 +143,30 @@ func TestIsRetryable(t *testing.T) {
 
 }
 
+func TestIsNotReady(t *testing.T) {
+	t.Run("plain error is not a not-ready condition", func(t *testing.T) {
+		assert.False(t, IsNotReady(fmt.Errorf("connection refused")))
+	})
+
+	t.Run("NotReadyError is a not-ready condition", func(t *testing.T) {
+		err := NewNotReadyError("deploy request has not staged its changes yet")
+		assert.True(t, IsNotReady(err))
+	})
+
+	t.Run("wrapped NotReadyError is a not-ready condition", func(t *testing.T) {
+		err := fmt.Errorf("cutover deploy request #7: %w", NewNotReadyError("not staged"))
+		assert.True(t, IsNotReady(err))
+	})
+
+	t.Run("NotReadyError stays retryable", func(t *testing.T) {
+		assert.True(t, IsRetryable(NewNotReadyError("not staged")))
+	})
+
+	t.Run("nil is not a not-ready condition", func(t *testing.T) {
+		assert.False(t, IsNotReady(nil))
+	})
+}
+
 func TestIsTransientTransportError(t *testing.T) {
 	tests := []struct {
 		name string
--- a/pkg/engine/errors.go
+++ b/pkg/engine/errors.go
@@ -20,6 +20,32 @@ func NewPermanentError(msg string, args ...any) error {
 	return &PermanentError{Err: fmt.Errorf(msg, args...)}
 }
 
+// NotReadyError wraps an error to indicate the engine's backend was not yet
+// ready to accept the operation and is expected to accept it once it catches
+// up. Polling drives should treat this as an expected condition and reattempt
+// on a later tick rather than reporting the attempt as a failure.
+type NotReadyError struct {
+	Err error
+}
+
+func (e *NotReadyError) Error() string { return e.Err.Error() }
+func (e *NotReadyError) Unwrap() error { return e.Err }
+
+// NewNotReadyError wraps err as a not-ready error.
+func NewNotReadyError(msg string, args ...any) error {
+	return &NotReadyError{Err: fmt.Errorf(msg, args...)}
+}
+
+// IsNotReady reports whether err indicates the engine's backend was not yet
+// ready to accept the operation.
+func IsNotReady(err error) bool {
+	if err == nil {
+		return false
+	}
+	var notReady *NotReadyError
+	return errors.As(err, &notReady)
+}
+
 // IsRetryable returns true if the error should be retried by the operator.
 // All errors are retryable by default, and engines explicitly wrap only
 // permanent errors with PermanentError.
--- a/pkg/engine/planetscale/control.go
+++ b/pkg/engine/planetscale/control.go
@@ -85,12 +85,24 @@ func (e *Engine) Start(ctx context.Context, req *engine.ControlRequest) (*engine
 func (e *Engine) Cutover(ctx context.Context, req *engine.ControlRequest) (*engine.ControlResult, error) {
 	return e.controlDeployRequest(ctx, engine.ControlCutover, req, "cutover", "Cutover initiated for",
 		func(ctx context.Context, client psclient.PSClient, number uint64) (*ps.DeployRequest, error) {
-			return client.ApplyDeployRequest(ctx, &ps.ApplyDeployRequestRequest{
+			dr, err := client.ApplyDeployRequest(ctx, &ps.ApplyDeployRequestRequest{
 				Organization: credOrg(req.Credentials),
 				Database:     req.Database,
 				Number:       number,
 			})
+			if err != nil && isDeployNotStagedError(err) {
+				return nil, engine.NewNotReadyError("deploy request has not staged its changes yet: %w", err)
+			}
+			return dr, err
 		})
+}
+
+// isDeployNotStagedError reports whether the PlanetScale API rejected a
+// cutover because the deploy request has not staged its changes yet. The API
+// briefly reports pending_cutover before the staged changes are visible to the
+// apply endpoint, so the rejection clears on its own once staging completes.
+func isDeployNotStagedError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "changes have not been staged")
 }
 
 // Revert rolls back a completed schema change during the revert window.
@@ -150,6 +162,12 @@ func (e *Engine) controlDeployRequest(
 
 	dr, err := mutate(ctx, client, meta.DeployRequestID)
 	if err != nil {
+		// A not-ready rejection means the deploy request exists and will accept
+		// the operation once its backend catches up — the deleted-deploy-request
+		// hint would misdirect whoever reads the message.
+		if engine.IsNotReady(err) {
+			return nil, fmt.Errorf("%s deploy request #%d: %w", errVerb, meta.DeployRequestID, err)
+		}
 		return nil, fmt.Errorf("%s deploy request #%d (may have been deleted): %w", errVerb, meta.DeployRequestID, err)
 	}
 
--- a/pkg/engine/planetscale/control_test.go
+++ b/pkg/engine/planetscale/control_test.go
@@ -1,12 +1,18 @@
 package planetscale
 
 import (
+	"context"
+	"errors"
+	"log/slog"
+	"os"
 	"testing"
 
+	ps "github.com/planetscale/planetscale-go/planetscale"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/block/schemabot/pkg/engine"
+	"github.com/block/schemabot/pkg/psclient"
 )
 
 func TestStop_UsesDeployRequestCancel(t *testing.T) {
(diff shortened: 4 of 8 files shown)


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/762

  **Verdict:** 2 findings — the not-ready classification is only consumed by
  the auto-cutover path, and the quiet retry loop has no escalation bound.

  ### Findings

  **1. CONFIRMED — the deferred/operator-initiated cutover path terminally
  fails the durable cutover request on the same self-clearing rejection.**
  [pkg/tern/local_control.go:294](https://github.com/block/schemabot/blob/043
  5a4939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_control.go#L294)

  `processPendingCutoverControlRequest` → [`c.cutover` →
  `eng.Cutover`](https://github.com/block/schemabot/blob/0435a4939fa49eeaa34c
  88d52f731ca50b59f075/pkg/tern/local_control.go#L164) returns the new
  `NotReadyError` (the `%w` chain survives, so `engine.IsNotReady(err)` would
  be `true` here), but this path never checks it: it logs a
  **LogLevelError** `"Cutover failed: ..."` apply event
  ([local_control.go:166-167](https://github.com/block/schemabot/blob/0435a49
  39fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_control.go#L166)) and
  calls `failPendingControlRequests`
  ([local_control.go:294](https://github.com/block/schemabot/blob/0435a4939fa
  49eeaa34c88d52f731ca50b59f075/pkg/tern/local_control.go#L294)), which is
  explicitly *terminal* ("A failed request is no longer pending, so the
  operator-owned retry loop stops re-running the operation",
  [control_requests.go:69-72](https://github.com/block/schemabot/blob/0435a49
  39fa49eeaa34c88d52f731ca50b59f075/pkg/tern/control_requests.go#L69)) — and
  it has already moved the apply to `cutting_over` at
  [local_control.go:285](https://github.com/block/schemabot/blob/0435a4939fa4
  9eeaa34c88d52f731ca50b59f075/pkg/tern/local_control.go#L285) before the
  attempt.

  *Failure scenario:* a `defer_cutover` apply reaches `waiting_for_cutover`
  the instant PlanetScale reports `pending_cutover`
  ([planetscale.go:772](https://github.com/block/schemabot/blob/0435a4939fa49
  eeaa34c88d52f731ca50b59f075/pkg/engine/planetscale/planetscale.go#L772));
  the user (or a queued request from earlier) triggers cutover on that first
  tick — exactly the not-staged window this PR documents. The engine returns
  NotReady, the durable request is marked FAILED with "changes have not been
  staged", an Error apply event lands in the user-visible timeline, and the
  drive owner exits ([local_apply_grouped.go:747-751](https://github.com/bloc
  k/schemabot/blob/0435a4939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_ap
  ply_grouped.go#L747)). The apply itself recovers (`cutting_over` is
  claimable), but in defer mode the auto-cutover branch is gated on
  `!opts.DeferCutover`
  ([local_apply_grouped.go:783](https://github.com/block/schemabot/blob/0435a
  4939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_grouped.go#L783)),
  so the cutover simply never happens until a human notices the failure and
  re-issues it — for the exact condition the PR's own `NotReadyError` doc
  comment says must not be reported as a failure. Fix mirrors the auto path:
  on `engine.IsNotReady(err)`, leave the control request pending and return
  nil so the next tick retries.

  **2. CONFIRMED — the quiet retry is unbounded with zero escalation; a
  genuinely wedged deploy request now stalls silently where it used to
  page.** [pkg/tern/local_apply_grouped.go:789](https://github.com/block/sche
  mabot/blob/0435a4939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_gr
  ouped.go#L789)

  The auto-cutover path has no bound of any kind: the
  `waitingForManualActionTimeout` (14d) at
  [local_apply_grouped.go:816](https://github.com/block/schemabot/blob/0435a4
  939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_grouped.go#L816)
  requires `opts.DeferCutover`, the `consecutiveErrors` counter tracks only
  `Progress` errors
  ([local_apply_grouped.go:658](https://github.com/block/schemabot/blob/0435a
  4939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_grouped.go#L658)),
  the drive context has no deadline, the only apply metric
  (`schemabot.active_applies`) has no state/age dimension, and the operator
  heartbeat keeps the lease fresh — masking the stall. Before this PR a
  permanently-unstaged deploy request (staging job wedged server-side while
  state keeps reporting `pending_cutover`) produced an Error log every 500ms
  tick ([ticker at
  local_apply_grouped.go:587](https://github.com/block/schemabot/blob/0435a49
  39fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_grouped.go#L587))
  that Error-log monitors could key on; now it produces Info forever. The
  substring classifier can't distinguish the seconds-long race from a
  permanently stuck request — they return the same message.

  *Failure scenario:* PlanetScale staging wedges after `pending_cutover` is
  reported; the apply sits in `waiting_for_cutover` indefinitely emitting
  only Info logs — zero operator signal where there used to be a page.
  Suggested fix: escalate to Warn/Error (or fall through to the `default:`
  arm) after N consecutive not-ready ticks or M minutes in the window, since
  the legitimate window lasts seconds.

  The whole fix rests on the quiet path actually retrying — if a NotReady
  rejection changed state or stopped the loop, downgrading the log would
  convert a visible failure into a lost cutover. Verified it does retry: the
  rejection changes no stored state, the engine's next `Progress` poll still
  maps `pending_cutover` → `StateWaitingForCutover`
  ([planetscale.go:772](https://github.com/block/schemabot/blob/0435a4939fa49
  eeaa34c88d52f731ca50b59f075/pkg/engine/planetscale/planetscale.go#L772)),
  and the branch at
  [local_apply_grouped.go:783](https://github.com/block/schemabot/blob/0435a4
  939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_grouped.go#L783)
  re-fires `eng.Cutover` on the next 500ms tick. Also verified the
  classification is fail-safe in the other direction: only the exact
  not-staged substring on the Cutover mutate maps to NotReady
  ([control.go:93-94](https://github.com/block/schemabot/blob/0435a4939fa49ee
  aa34c88d52f731ca50b59f075/pkg/engine/planetscale/control.go#L93)), every
  other error keeps pre-PR Error behavior, and `NotReadyError` stays
  retryable (test-asserted), so no genuine failure is swallowed.

  ### Notes (not blocking)

  - **Event-log hygiene (pre-existing, but this PR blesses the loop):**
  `LogEventCutoverTriggered` ("Auto-triggering cutover") is written to
  storage *before* every attempt, every 500ms, with no one-shot guard
  ([local_apply_grouped.go:785](https://github.com/block/schemabot/blob/0435a
  4939fa49eeaa34c88d52f731ca50b59f075/pkg/tern/local_apply_grouped.go#L785)),
  and the NotReady arm writes no apply event — so the user-visible timeline
  shows repeated triggers with no outcome (~2-20 duplicate rows for the
  normal seconds-long window). `atomicPollState` already has one-shot flags
  for exactly this shape (`revertSkipped`, `warnedPerShardUnavailable`);
  consider one here.
  - **Unpinned server prose:** `isDeployNotStagedError`
  ([control.go:104](https://github.com/block/schemabot/blob/0435a4939fa49eeaa
  34c88d52f731ca50b59f075/pkg/engine/planetscale/control.go#L104)) does a
  case-sensitive `strings.Contains` on raw PlanetScale API copy that appears
  nowhere in planetscale-go. It follows package precedent
  (`isSnapshotInProgress`) and fails safe (reverts to pre-PR Error behavior),
  but a PS rewording silently disables the fix with no signal.

  ### Verified correct

  - Go mechanics: `NotReadyError.Unwrap`/`errors.As` chain survives both `%w`
  rewraps (engine → `controlDeployRequest` → tern); `IsNotReady(nil)` is
  false; the switch-with-initializer with an empty `case err == nil:` arm is
  correct Go and preserves the success path.
  - `Start`/`Revert`/`Stop` behavior unchanged: only the Cutover mutate can
  produce NotReady, and the `controlDeployRequest` branch at
  [control.go:168](https://github.com/block/schemabot/blob/0435a4939fa49eeaa3
  4c88d52f731ca50b59f075/pkg/engine/planetscale/control.go#L168) only alters
  message text (drops the misleading "may have been deleted" hint — nothing
  downstream parses that string; grep confirmed).
  - `NotReadyError` composes correctly with the existing classification:
  stays retryable (only `PermanentError` is not), so the Progress-side
  consecutive-error/permanent-failure machinery is unaffected.
  - gRPC/remote topology still gets the fix: the Cutover RPC handler only
  durably queues (never invokes the engine,
  [local_control.go:23-25](https://github.com/block/schemabot/blob/0435a4939f
  a49eeaa34c88d52f731ca50b59f075/pkg/tern/local_control.go#L23)); the
  engine's rejection is classified in-process inside the remote data plane's
  own drive loop, so the typed error never needs to cross the wire.
  - Altitude is right: pre-checking `DeployRequest.DeploymentState` before
  cutover would TOCTOU-race — the state already reads `pending_cutover`
  inside the window — so classifying the rejection is the correct depth.
  - New tests match package conventions (`NewWithClient` fake client,
  `t.Context()`, testify) and cover both branches: not-staged → NotReady
  without the deleted hint; other errors → plain error with the hint. CI
  fully green (unit, integration, all E2E suites including the Vitess
  matrix).

Kiran01bm (approved):

aparajon:
🤖 Both findings addressed in 810b89e2:

1. **Durable cutover path:** `processPendingCutoverControlRequest` now classifies `engine.IsNotReady` after `c.cutover` and leaves the request pending (Info log) so the next progress tick retries it, instead of terminally failing the command. The rejected attempt also no longer writes the Error "Cutover failed" apply event — the trigger event is recorded only for the attempt the backend accepts or definitively fails.
2. **Escalation bound:** the auto-cutover arm (extracted to `autoTriggerCutover`) tracks the first not-ready rejection in `atomicPollState`; a rejection persisting past `cutoverNotReadyEscalationAfter` (2m) records a one-time Error timeline event and Error-logs every further tick, restoring the pageable signal for a wedged deploy request. An accepted cutover resets the clock.

Note A: took the suggestion — `cutoverTriggerLogged` one-shots the auto trigger event per drive. Note B: left as-is; the substring follows package precedent (`isSnapshotInProgress`) and fails safe — a PlanetScale rewording reverts to pre-PR Error behavior, never loses a cutover.

Covered by `TestLocalClient_ProcessPendingCutoverControlRequestRetriesWhenCutoverNotReady` and `TestLocalClient_AutoTriggerCutoverEscalatesWhenNotReadyPersists`.

*This reply was generated by Claude Code (claude-fable-5).*

PR #764, opened 2026-07-18
fix(github): show the revert-window countdown for remote applies

aparajon:
## Why this matters

For applies dispatched to a remote data plane, the PR status comment showed **Status: Revert Window** with no "Closes in Xm Ys" countdown, so the operator had no idea how long they had left to revert. Locally-driven applies were unaffected, which made the gap easy to miss: the engine stamps the deadline into its resume metadata and projects it into the progress response's display map, but the control-plane mirror that persists remote display metadata dropped `revert_expires_at` on the floor.

The per-table line for that state also overstated finality: "✓ Complete (pending revert)" reads as done-and-final while the change is still revertable, and "(pending revert)" reads as if a revert is queued when nothing happens unless the operator acts.

## What it does

- `PSDisplayMetadataStorageBlob` now carries `revert_expires_at` through the remote display mirror into the stored projection the comment renders from, and treats the deadline alone as worth storing. A malformed timestamp is a hard error rather than a silently missing countdown.
- The revert-window table line renders as **Complete (revert window open)** — no checkmark, so ✅ stays reserved for the final completed state; the header's `Revert Window | Closes in Xm Ys` carries the countdown. The CLI progress view renders the same line. The genuinely-completed table line uses ✅ to match the comment's other glyphs.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -33,7 +33,7 @@ fi
 scripts/lint-fix.sh
 
 # Regenerate TEMPLATES.md if any template source files changed
-TEMPLATE_FILES=$(git diff --cached --name-only -- 'pkg/webhook/templates/*.go' 'pkg/cmd/templates/*.go')
+TEMPLATE_FILES=$(git diff --cached --name-only -- 'pkg/webhook/templates/*.go' 'pkg/cmd/internal/templates/*.go')
 if [ -n "$TEMPLATE_FILES" ]; then
     # Ensure Go is in PATH (hooks may not inherit the full user PATH)
     export PATH="$PATH:$HOME/go/bin:/usr/local/go/bin:/opt/homebrew/bin"
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -2858,19 +2858,19 @@ Deploy request: https://app.planetscale.com/acme/myapp/deploy-requests/42
 
 **Keyspace `testapp`**
 
-**`orders`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✅ Complete (pending revert)
+**`orders`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
 
 ```sql
 ALTER TABLE `orders` ADD INDEX `idx_user_id`(`user_id`);
 ```
 
-**`users`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✅ Complete (pending revert)
+**`users`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
 
 ```sql
 ALTER TABLE `users` ADD INDEX `idx_email`(`email`);
 ```
 
-**`products`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✅ Complete (pending revert)
+**`products`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
 
 ```sql
 ALTER TABLE `products` ADD INDEX `idx_price`(`price_cents`);
@@ -2909,19 +2909,19 @@ Deploy request: https://app.planetscale.com/acme/myapp/deploy-requests/42
 
 **Keyspace `testapp`**
 
-**`orders`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✅ Complete (pending revert)
+**`orders`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
 
 ```sql
 ALTER TABLE `orders` ADD INDEX `idx_user_id`(`user_id`);
 ```
 
-**`users`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✅ Complete (pending revert)
+**`users`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
 
 ```sql
 ALTER TABLE `users` ADD INDEX `idx_email`(`email`);
 ```
 
-**`products`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✅ Complete (pending revert)
+**`products`**: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
 
 ```sql
 ALTER TABLE `products` ADD INDEX `idx_price`(`price_cents`);
@@ -4966,7 +4966,7 @@ Vitess plan: Multi-keyspace with DDL + VSchema across keyspaces
 
   ── myapp_sharded ──
 
-     ~ orders: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 ✓ Complete (pending revert)
+     ~ orders: 🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨 Complete (revert window open)
        ALTER TABLE `orders` ADD INDEX `idx_total`(`total_cents`);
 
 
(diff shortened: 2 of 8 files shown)


Kiran01bm:
🤖 Adversarial correctness review (posted by Kiran's adversarial review AI
  agent).

  **Verdict: clean — approve.** (No blocking findings; a few non-blocking
  notes below, mostly about two test comments that document behavior the code
  doesn't have.)

  The relaxed "nothing worth storing" guard
  ([state_converters.go:441](https://github.com/block/schemabot/blob/370d8b79
  5dc7b9812f61999701569f5a7bd1d1ba/pkg/tern/state_converters.go#L441) now
  keeps blobs containing only the deadline) combined with the mirror's
  replace-not-merge write (`SaveEngineResumeState` does a wholesale `SET
  ao.engine_resume_metadata = ?`, [apply_operations.go:542](https://github.co
  m/block/schemabot/blob/370d8b795dc7b9812f61999701569f5a7bd1d1ba/pkg/storage
  /mysqlstore/apply_operations.go#L542); `lastBlob` is only a dedup, never a
  merge source) means a display map carrying *only* `revert_expires_at` would
  clobber the previously mirrored deploy-request URL exactly when the
  operator needs it to revert — and the new test
  `TestPSDisplayMetadataStorageBlobRevertExpiresOnly` explicitly blesses that
  input.

  Verified it cannot happen: the served display map has exactly one producer
  — `LocalClient.Progress` → `loadStoredDisplayMetadata` →
  `PSDisplayMetadata` over the operation's stored engine blob
  ([local_client.go:2479](https://github.com/block/schemabot/blob/370d8b795dc
  7b9812f61999701569f5a7bd1d1ba/pkg/tern/local_client.go#L2479)), with no
  bypass. `revert_expires_at` enters that blob only via
  `setRevertExpiresAtMetadata`
  ([local_apply_grouped.go:687](https://github.com/block/schemabot/blob/370d8
  b795dc7b9812f61999701569f5a7bd1d1ba/pkg/tern/local_apply_grouped.go#L687)),
  gated on `StateRevertWindow`, whose sole producer is the PlanetScale
  engine's `pending_revert` mapping — which requires `DeployRequestID != 0`,
  and every `psMetadata` persisted with a `DeployRequestID` also carries
  `BranchName` and `DeployRequestURL` (deploy-request creation, deferred
  deploy, and post-deploy persists all stamp both; the merge is
  key-preserving; the JSON decode is all-or-nothing). So every map with the
  deadline is a superset of what's already stored, and the mirror's
  revert-window write stores the URL + branch + deadline together — the shape
  the updated `TestGRPCClientMirrorsRemoteDisplayMetadata` proves survives.

  ### Notes (not blocking)

  - **Two test comments document behavior the code doesn't have.** (1) The
  comment above [`TestPSDisplayMetadataStorageBlobBadRevertExpires`](https://
  github.com/block/schemabot/blob/370d8b795dc7b9812f61999701569f5a7bd1d1ba/pk
  g/tern/state_converters_test.go#L168) claims the mirror "retries on the
  next poll with the rest of the display state intact" — in fact the
  malformed value comes from the data plane's durable metadata, so every
  retry fails identically, and `mirrorRemoteDisplayMetadata` persists
  *nothing* from the failing poll
  ([grpc_client.go:1424-1428](https://github.com/block/schemabot/blob/370d8b7
  95dc7b9812f61999701569f5a7bd1d1ba/pkg/tern/grpc_client.go#L1424)): "intact"
  only means a previously stored blob survives frozen; if the first-ever
  poll is malformed, nothing renders at all. (2) The comment above
  [`TestPSDisplayMetadataStorageBlobRevertExpiresOnly`](https://github.com/bl
  ock/schemabot/blob/370d8b795dc7b9812f61999701569f5a7bd1d1ba/pkg/tern/state_
  converters_test.go#L154) claims a revert-window progress response "may
  carry only revert_expires_at" — no code path can produce that map (see
  above), and the comment launders an input the merge-free mirror could not
  store safely. Both comments should be corrected.
  - **Hard-error scope.** A malformed `revert_expires_at` fails the *entire*
  blob ([state_converters.go:427](https://github.com/block/schemabot/blob/370
  d8b795dc7b9812f61999701569f5a7bd1d1ba/pkg/tern/state_converters.go#L427)),
  suppressing the deploy-request URL, branch, and VSchema mirror too — the
  opposite of the same pipeline's stated policy one field over ("A malformed
  VSchema blob should not also drop the deploy-request URL, so log and
  continue", [webhook/apply.go:93-94](https://github.com/block/schemabot/blob
  /370d8b795dc7b9812f61999701569f5a7bd1d1ba/pkg/webhook/apply.go#L93)) and of
  both renderers, which skip an unparseable deadline. Verified the error
  branch is effectively unreachable today: every writer since the key's
  introduction emits `UTC().Format(time.RFC3339)`, and the wire value is
  re-formatted from a parsed `*time.Time` on every read, so even stored-blob
  corruption degrades gracefully upstream. Fine to keep as a version-skew
  tripwire, but log-and-skip-the-field would match the pipeline's own
  precedent.
  - **Glyph consistency.** Completed table lines are now ✅ but the
  shard-bucket summary in the same comment still renders `N ✓` ([webhook/temp
  lates/apply.go:800](https://github.com/block/schemabot/blob/370d8b795dc7b98
  12f61999701569f5a7bd1d1ba/pkg/webhook/templates/apply.go#L800),
  [:828](https://github.com/block/schemabot/blob/370d8b795dc7b9812f6199970156
  9f5a7bd1d1ba/pkg/webhook/templates/apply.go#L828)), and the CLI keeps `✓
  Complete` / `✓ Complete (finalizing)`
  ([progress.go:447](https://github.com/block/schemabot/blob/370d8b795dc7b981
  2f61999701569f5a7bd1d1ba/pkg/cmd/internal/templates/progress.go#L447),
  [:557](https://github.com/block/schemabot/blob/370d8b795dc7b9812f6199970156
  9f5a7bd1d1ba/pkg/cmd/internal/templates/progress.go#L557)). Cosmetic.
  - **Pre-existing tooling gap.** The pre-commit hook's TEMPLATES.md
  regeneration trigger globs `pkg/cmd/templates/*.go` — a directory that
  doesn't exist; the CLI templates live at `pkg/cmd/internal/templates/`
  ([.githooks/pre-commit:36](https://github.com/block/schemabot/blob/370d8b79
  5dc7b9812f61999701569f5a7bd1d1ba/.githooks/pre-commit#L36)) — and no CI job
  checks golden freshness. This PR's goldens are fresh only because the
  webhook template also changed; a future CLI-only template edit will
  silently skip regeneration.

  ### Verified correct

  - **TEMPLATES.md goldens are fresh**: `scripts/update-templates.sh` re-run
  at the PR head produced zero diff; no stale "✓ Complete (pending revert)"
  or "(pending revert)" strings remain anywhere in Go code or goldens.
  - **Round-trip format stability**: the producer emits
  `UTC().Format(time.RFC3339)` ([planetscale/progress.go:630](https://github.
  com/block/schemabot/blob/370d8b795dc7b9812f61999701569f5a7bd1d1ba/pkg/engin
  e/planetscale/progress.go#L630)), the storage projection re-emits the same
  shape ([state_converters.go:379](https://github.com/block/schemabot/blob/37
  0d8b795dc7b9812f61999701569f5a7bd1d1ba/pkg/tern/state_converters.go#L379)),
  and the blob-equality dedup is unaffected (deterministic marshal per
  input).
  - **The mirror actually runs during the revert window**: all remote drive
  paths funnel into the single `pollForCompletion` loop, which mirrors
  display metadata before any state gating and keeps polling through the
  non-terminal `revert_window` state — so the deadline reaches storage while
  the countdown matters.
  - **The countdown doesn't freeze**: the comment observer re-renders every
  30s during a stagnant revert window, and both consumers (webhook header
  `revertWindowCountdown`, CLI progress view) read the same
  `revert_expires_at` key the mirror stores; the API overlay copies all
  display keys, so the CLI countdown works from stored state for remote
  applies too.
  - **Blob replacement zeroing engine-only fields (`DeployRequestID`,
  `DeployedAt`) is harmless**: the control plane never ships stored metadata
  back to the data plane; the mirror write preserves `EngineResumeContext`.
  - **No string-parsing consumers of the reworded lines**: comment updates
  are re-rendered and PATCHed wholesale, so the one-time rewrite of open PR
  comments on deploy is benign; the `strings.Count(result, "Complete") == 2`
  assertion remains valid.
  - **Version skew is graceful**: an older data plane simply never emits
  `revert_expires_at` (missing countdown, pre-PR behavior); an older control
  plane ignores the extra key.
  - CI fully green (unit, integration, all E2E suites including the Vitess
  matrix and gRPC multi-deployment).

Kiran01bm (approved):

aparajon:
🤖 Thanks for the verification pass — the map of the deadline-only blob's provenance chain was especially useful. Notes addressed in 0dc30634:

**Test comments** — both corrected. The revert-expires-only test now says what it actually pins: the "nothing worth storing" guard counts `revert_expires_at` as a display field in its own right, and live progress responses always carry the deploy-request URL alongside the deadline (so the map the old comment described has no producer). The malformed-deadline test now states that the conversion persists nothing from a failing poll and a previously stored blob simply survives as-is — no "retries with the rest intact" claim.

**Hard-error scope** — kept as the version-skew tripwire, per your note that that's fine. Since every writer emits `UTC().Format(time.RFC3339)` and the wire value is re-formatted from a parsed `*time.Time` on every read, a malformed deadline can only mean skew or corruption — degrading around it would hide exactly the signal worth surfacing. The corrected test comment now documents this policy explicitly.

**Glyph consistency** — leaving as-is. The shard-bucket line's `✓ ◐ ● ⏳ ✗ …` set is a deliberately compact single-cell vocabulary for the dense summary line; swapping only `✓` for the emoji `✅` would break that set's own internal consistency (`✗`, `◐`, `●` have no emoji peers in it). The CLI's `✓ Complete` is terminal styling, consistent within the CLI.

**Pre-commit glob** — fixed in the same commit: the TEMPLATES.md trigger now globs `pkg/cmd/internal/templates/*.go`, so CLI-only template edits regenerate the goldens.

*This reply was generated by Claude Code (claude-fable-5).*

PR #770, opened 2026-07-19
fix(api): reclaim stale reverting and skipping_revert applies

aparajon:
## Why this matters

An apply dwells in `reverting` / `skipping_revert` while the engine reverts or finalizes the deploy request. If the driver dies there — routine pod churn is enough — the apply is stranded non-terminal forever: the stale-heartbeat re-claim is the only recovery path after a driver death, and neither revert-phase state was in `claimableApplyStates()`. Every operation-loop claim then logs `parent apply not claimable for operation; operation will be retried` indefinitely, the revert never resolves, and the progress comment freezes.

```
driver dies mid-revert (pod churn)
        │
        ▼
apply: reverting, heartbeat stale ──► re-claim arm: state IN (claimable) AND stale
                                        └─ now includes reverting / skipping_revert
        │
        ▼
peer re-claims, re-attaches to the in-flight engine revert, drives to terminal
```

## What it does

- Adds `reverting` and `skipping_revert` to `claimableApplyStates()`. The claim only rotates the lease — the persisted state is untouched, so the resumed drive re-attaches to the in-flight engine revert; the durable revert / skip-revert signals mean a re-claim never re-issues the command.
- Keeps revert-phase tasks active through the resume re-plan. The re-plan completes tasks whose table has dropped out of the live-schema diff, but post-cutover the live schema matches the reviewed target *by definition* while the engine holds the revert window open or unwinds a revert — so a reclaim there would have terminalized the apply as a success while the schema change was being reverted underneath it. Revert-phase tasks now stay active so the resume re-attaches to the engine and their terminal state comes from engine progress, never from a schema comparison.
- Preserves revert-phase task and apply states through the post-reattach persistence step, so the revert-phase marker survives repeated driver deaths instead of being durably rewritten to `running` before the first progress tick.
- Fails the grouped resume closed when a revert-phase Vitess apply has no persisted engine resume state — a fresh engine apply there would re-deploy the reviewed schema change on top of the revert, so recovery retries against intact storage instead.
- Leaves terminal `reverted` tasks terminal through the resume re-plan, and terminalizes an all-terminal apply with the state derived from its task states — all-reverted finishes as `reverted`, never as a false `completed`.
- Documents the invariant on the claimable list: every non-terminal state an apply can dwell in must be re-claimable (or have its own claim arm), or any pod restart orphans it.
- Tests prove stale revert-phase applies are reclaimable via both `FindNextApply` and `ClaimApplyByID` (the operation-loop path), that a fresh lease is never stolen, that the claim preserves the revert-phase state, that the resume re-plan keeps revert-phase tasks active whether or not the revert has landed, that reattach persistence preserves revert-phase states while still projecting forward states, that a revert-phase resume without persisted engine state fails closed, and that an all-reverted resume terminalizes as `reverted`.

### DIFF
--- a/integration/operator_test.go
+++ b/integration/operator_test.go
@@ -879,6 +879,8 @@ func TestOperator_ClaimableStates(t *testing.T) {
 		{name: "waiting for cutover", applyState: state.Apply.WaitingForCutover, databaseType: "vitess", engine: "planetscale", wantClaim: true},
 		{name: "cutting over", applyState: state.Apply.CuttingOver, databaseType: "vitess", engine: "planetscale", wantClaim: true},
 		{name: "revert window", applyState: state.Apply.RevertWindow, databaseType: "vitess", engine: "planetscale", wantClaim: true},
+		{name: "reverting", applyState: state.Apply.Reverting, databaseType: "vitess", engine: "planetscale", wantClaim: true},
+		{name: "skipping revert", applyState: state.Apply.SkippingRevert, databaseType: "vitess", engine: "planetscale", wantClaim: true},
 		{name: "completed", applyState: state.Apply.Completed, databaseType: "mysql", engine: "spirit"},
 		{name: "failed", applyState: state.Apply.Failed, databaseType: "mysql", engine: "spirit"},
 		{name: "stopped", applyState: state.Apply.Stopped, databaseType: "mysql", engine: "spirit"},
--- a/pkg/storage/mysqlstore/applies.go
+++ b/pkg/storage/mysqlstore/applies.go
@@ -93,8 +93,10 @@ type txBeginner interface {
 // apply can dwell in them while the engine works: a driver killed by routine
 // pod churn mid-revert must not strand the apply non-terminal forever. Recovery
 // re-attaches to the engine (the deploy request keeps reverting or finalizing
-// server-side) and drives the apply to its terminal state; the durable
-// revert/skip-revert signals mean a re-claim never re-issues the command.
+// server-side) and drives the apply to its terminal state. The durable
+// revert/skip-revert signals mean a re-claim normally does not re-issue the
+// command; a crash between the engine accepting the command and the durable
+// write can cause a harmless re-issue, which the engine treats as idempotent.
 //
 // Every non-terminal state an apply can dwell in must appear here (or have its
 // own claim arm, like pending and stopped-with-start): the stale-heartbeat
--- a/pkg/storage/mysqlstore/applies.go
+++ b/pkg/storage/mysqlstore/applies.go
@@ -88,6 +88,18 @@ type txBeginner interface {
 // holds while waiting for the data plane to leave stopped after a start. A
 // stale heartbeat there means the driver died mid-resume, and recovery must be
 // able to reclaim and finish driving it to a terminal state.
+//
+// The revert-phase states (reverting, skipping_revert) are included because an
+// apply can dwell in them while the engine works: a driver killed by routine
+// pod churn mid-revert must not strand the apply non-terminal forever. Recovery
+// re-attaches to the engine (the deploy request keeps reverting or finalizing
+// server-side) and drives the apply to its terminal state; the durable
+// revert/skip-revert signals mean a re-claim never re-issues the command.
+//
+// Every non-terminal state an apply can dwell in must appear here (or have its
+// own claim arm, like pending and stopped-with-start): the stale-heartbeat
+// re-claim is the only recovery path after a driver dies, and a dwellable state
+// missing from this list is orphaned by any pod restart.
 func claimableApplyStates() []string {
 	return []string{
 		state.Apply.Running,
@@ -99,6 +111,8 @@ func claimableApplyStates() []string {
 		"recovering_cutover",
 		state.Apply.CuttingOver,
 		state.Apply.RevertWindow,
+		state.Apply.Reverting,
+		state.Apply.SkippingRevert,
 		state.Apply.PreparingBranch,
 		state.Apply.ApplyingBranchChanges,
 		state.Apply.ValidatingBranch,
--- a/pkg/storage/mysqlstore/applies_test.go
+++ b/pkg/storage/mysqlstore/applies_test.go
@@ -2017,6 +2017,100 @@ func TestApplyStore_ClaimApplyByIDClaimsStaleSetupPhase(t *testing.T) {
 	assert.NotEmpty(t, claimed.LeaseToken)
 }
 
+// An apply dwells in a revert-phase state (reverting, skipping_revert) while
+// the engine reverts or finalizes the deploy request. If the driver dies there
+// — routine pod churn is enough — a fresh driver must reclaim the apply once
+// the heartbeat goes stale and resume polling the engine to its terminal
+// state; otherwise the apply is stranded non-terminal forever with no driver.
+// While the original driver is alive its heartbeat stays fresh, so the apply
+// is never claimed out from under it.
+func TestApplyStore_FindNextApplyClaimsStaleRevertPhase(t *testing.T) {
+	for _, revertState := range []string{
+		state.Apply.Reverting,
+		state.Apply.SkippingRevert,
+	} {
+		t.Run(revertState, func(t *testing.T) {
+			clearTables(t)
+			ctx := t.Context()
+			store := New(testDB)
+
+			lock := createTestLock(t, store, "testdb", storage.DatabaseTypeVitess, "staging")
+			apply := createTestApplyWithStateAndEnv(t, store, lock, "apply_revert_"+revertState, 702, revertState, "staging")
+			require.Equal(t, storage.EnginePlanetScale, apply.Engine, "revert-phase states only occur for the PlanetScale engine")
+
+			fresh, err := store.Applies().FindNextApply(ctx, "operator-a")
+			require.NoError(t, err)
+			assert.Nil(t, fresh, "a revert-phase apply with a fresh heartbeat is owned by its active driver")
+
+			_, err = testDB.ExecContext(ctx, `
+				UPDATE applies
+				SET updated_at = NOW() - INTERVAL 2 MINUTE
+				WHERE id = ?
+			`, apply.ID)
+			require.NoError(t, err)
+
+			claimed, err := store.Applies().FindNextApply(ctx, "operator-a")
+			require.NoError(t, err)
+			require.NotNil(t, claimed, "a stale revert-phase apply must be reclaimable for recovery")
+			assert.Equal(t, apply.ApplyIdentifier, claimed.ApplyIdentifier)
+			assert.Equal(t, revertState, claimed.State, "the caller sees the revert-phase state to resume from")
+			assert.Equal(t, "operator-a", claimed.LeaseOwner)
+			assert.NotEmpty(t, claimed.LeaseToken)
+		})
+	}
+}
+
+// The operation-level claim loop leases the orphaned apply_operations row
+// first and then needs the parent apply lease via ClaimApplyByID, so a stale
+// revert-phase parent must be reclaimable through it as well. The claim only
+// rotates the lease: the persisted state must stay the revert-phase state so
+// the resumed drive re-attaches to the in-flight engine revert instead of
+// restarting the apply.
+func TestApplyStore_ClaimApplyByIDClaimsStaleRevertPhase(t *testing.T) {
+	for _, revertState := range []string{
+		state.Apply.Reverting,
+		state.Apply.SkippingRevert,
+	} {
+		t.Run(revertState, func(t *testing.T) {
+			clearTables(t)
+			ctx := t.Context()
+			store := New(testDB)
+
+			lock := createTestLock(t, store, "testdb", storage.DatabaseTypeVitess, "staging")
+			apply := createTestApplyWithStateAndEnv(t, store, lock, "apply_claim_revert_"+revertState, 703, revertState, "staging")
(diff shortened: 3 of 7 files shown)


Kiran01bm:
"🤖 Review findings - created by Kiran's code review agent - for pull/770"

  **Verdict:** 2 findings — the fix works on the single-crash path it targets,
  but the resume itself durably erases the revert-phase states the new guard
  keys on, so a second driver death re-opens the same false-success one write
  later; plus a fail-open resume-state branch that a revert-phase resume newly
  reaches.

  ### Findings

  **1. [CONFIRMED] `launchAtomicResume` stomps reclaimed revert-phase
  tasks/apply to `running`, re-opening the false-success hole on a second
  driver death.**
  [pkg/tern/local_control_resume.go#L778-L801](https://github.com/block/schema
  bot/blob/9e934684bc4556cbab877ad0609e68dfd23f530f/pkg/tern/local_control_res
  ume.go#L778-L801) — right after `eng.Apply` re-attaches, the resume persists
  every kept-active task to `state.Task.Running` and the apply to
  `state.Apply.Running`; the only carve-out is `recovering`:

  ```go
  taskState := state.Task.Running
  if recovering {
      taskState = state.Task.Recovering
  }
  ```

  `reverting`/`skipping_revert` fall through, and `transitionTaskState` writes
  unconditionally — no state-machine refusal. This durably erases exactly the
  marker `taskInRevertPhase`
  ([#L500-L502](https://github.com/block/schemabot/blob/9e934684bc4556cbab877a
  d0609e68dfd23f530f/pkg/tern/local_control_resume.go#L500-L502)) keys on,
  *before* the first progress poll (500ms ticker; up to ~10 tolerated
  consecutive poll errors extend the window to minutes).

  Concrete failure: apply in `reverting`; driver 1 dies (this PR's own
  scenario). Driver 2 reclaims, the guard correctly keeps tasks active,
  reattach succeeds, `running`/`running` is persisted — then driver 2 dies
  before the first successful progress tick. Driver 3 reclaims a `running`
  apply with `running` tasks: `taskInRevertPhase` is false, post-cutover live
  schema matches the reviewed target while the revert hasn't landed, every
  task takes the `!stillNeeded` completion branch, and the apply terminalizes
  as `completed` while PlanetScale reverts the schema underneath — the exact
  false success the guard's comment says must never happen.

  Framing: the PR strictly improves the one-crash case (pre-PR there was no
  guard at all, so a single `revert_window` reclaim already false-completed).
  This is the residual two-crash path, and it defeats the guard for
  `revert_window` and the two newly claimable states alike. Fix shape:
  preserve revert-phase task states in the post-reattach transition loop
  (mirroring how `markApplyRecovering` preserves terminal task states), or
  derive the persisted post-reattach state from the engine resume state.

  **2. [PLAUSIBLE] `ErrEngineResumeStateNotFound` is the lone fail-open branch
  on the Vitess resume path, and a revert-phase resume that hits it launches
  a duplicate forward schema change.**
  [pkg/tern/local_control_resume.go#L879-L891](https://github.com/block/schema
  bot/blob/9e934684bc4556cbab877ad0609e68dfd23f530f/pkg/tern/local_control_res
  ume.go#L879-L891) returns a context-only `ResumeState` (empty `Metadata`) on
  not-found, while every adjacent failure (storage read error, unresolvable
  operation ID, post-accept save failure) fails closed via
  `errGroupedResumeStateUnavailable`. The PlanetScale engine takes the
  reattach path only when `Metadata != ""` ([pkg/engine/planetscale/apply.go#L
  80-L82](https://github.com/block/schemabot/blob/9e934684bc4556cbab877ad0609e
  68dfd23f530f/pkg/engine/planetscale/apply.go#L80-L82)); with empty metadata
  it runs the fresh path — new branch, forward DDL re-applied, new deploy
  request — with no existing-deploy-request lookup. Revert landed → the
  reviewed change is silently re-deployed, un-reverting a user-requested
  revert; revert in flight → `no_changes` result completes the apply as a
  success while the revert unwinds it.

  Why PLAUSIBLE not CONFIRMED: the row is written-or-fail-closed before the
  apply is ever marked running, and no deletion path exists, so an apply in
  revert phase can't lose its row through any code path — only storage row
  loss or a legacy-binary apply. But an apply dwelling in revert phase *by
  definition* has in-flight engine work, so a missing row there can never
  legitimately mean "start fresh"; this branch should fail closed for
  revert-phase (or all post-cutover) applies. Pre-PR this was unreachable for
  revert-phase resumes (the replan auto-completed them before
  `groupedResumeState` ran), so the PR does newly route traffic at it.

  Whether a reclaimed mid-revert apply can re-issue the reviewed schema change
  or the revert command. Proved safe on the normal path: the resume-state row
  is guaranteed present (save-or-fail-closed at launch, no deletion path), so
  `eng.Apply` routes to `resumeExistingDeployRequest`, whose only re-deploy
  gate is `dr.DeploymentState == deployState.Ready && !meta.DeferredDeploy &&
  dr.DeployedAt == nil`
  ([apply.go#L1015-L1017](https://github.com/block/schemabot/blob/9e934684bc45
  56cbab877ad0609e68dfd23f530f/pkg/engine/planetscale/apply.go#L1015-L1017)) —
  `in_progress_revert`/`complete_pending_revert`/`complete_revert` all take
  the reattach-only path. Skip-revert is re-seeded from durable
  `RevertSkippedAt`, and task-state coverage of the guard is complete
  (`skipping_revert` applies hold `revert_window` tasks; there is no
  task-level skip state to miss). The two findings above are the only cracks:
  both require an additional failure (second crash, or storage row loss) on
  top of the reclaim.

  ### Notes (not blocking)

  - **Doc overclaim** —
  [applies.go#L91-L102](https://github.com/block/schemabot/blob/9e934684bc4556
  cbab877ad0609e68dfd23f530f/pkg/storage/mysqlstore/applies.go#L91-L102): "the
  durable revert/skip-revert signals mean a re-claim never re-issues the
  command" is literally false for skip-revert — all three issuers call
  `eng.SkipRevert` *before* the durable write, and `markRevertSkipped`
  ([local_apply_grouped.go#L1109-L1113](https://github.com/block/schemabot/blo
  b/9e934684bc4556cbab877ad0609e68dfd23f530f/pkg/tern/local_apply_grouped.go#L
  1109-L1113)) is Warn-and-continue. The re-issue is harmless in behavior;
  soften the comment ("normally does not re-issue; a crash between
  engine-accept and the durable write can cause a harmless re-issue").
  - **Duplicate `database` slog key** — [local_control_resume.go#L462-L463](ht
  tps://github.com/block/schemabot/blob/9e934684bc4556cbab877ad0609e68dfd23f53
  0f/pkg/tern/local_control_resume.go#L462-L463): `append(task.LogAttrs(),
  "database", apply.Database)` duplicates the `"database"` key
  `Task.LogAttrs()` already emits
  ([logattrs.go#L81](https://github.com/block/schemabot/blob/9e934684bc4556cba
  b877ad0609e68dfd23f530f/pkg/storage/logattrs.go#L81)); empirically the new
  test logs two conflicting `database` attrs. Violates the repo's LogAttrs
  convention; drop the append.
  - **Half-vacuous test matrix** — [local_control_resume_test.go#L230](https:/
  /github.com/block/schemabot/blob/9e934684bc4556cbab877ad0609e68dfd23f530f/pk
  g/tern/local_control_resume_test.go#L230): the two "revert already landed"
  subtests pass with the guard deleted — `alterUsersEmailPlan()` DDL is
  byte-identical to `reviewed`, so the pre-existing else-branch yields the
  same observable state (active, same DDL, count 0). Use re-plan DDL that
  differs from `reviewed` (the discriminating input: without the guard it
  trips `verifyReplannedTaskDDL`).
  - **Terminal `reverted` tasks re-enter the replan** — the loop skips only
  `completed` ([local_control_resume.go#L449](https://github.com/block/schemab
  ot/blob/9e934684bc4556cbab877ad0609e68dfd23f530f/pkg/tern/local_control_resu
  me.go#L449)); a crash between the task sync and the apply CAS leaves
  tasks=`reverted`/apply=`reverting`, and the reclaim flips the reverted task
  to `running` for one poll tick before the engine re-terminalizes it as
  `reverted`. Verified self-healing (reattach-only, no re-deploy) — hygiene:
  skip terminal `reverted` tasks the way `completed` is skipped.
  - **Pre-existing: cancel-mid-revert falsely terminalizes** —
  `firstRevertWindowTask` checks only `revert_window`
  ([local_control.go#L1026](https://github.com/block/schemabot/blob/9e934684bc
  4556cbab877ad0609e68dfd23f530f/pkg/tern/local_control.go#L1026)) and
  `hasLiveEngineWork` omits `reverting`
  ([#L1098-L1106](https://github.com/block/schemabot/blob/9e934684bc4556cbab87
  7ad0609e68dfd23f530f/pkg/tern/local_control.go#L1098-L1106)), so a pending
  cancel processed while tasks are `reverting` marks them `cancelled` with no
  engine call. Reachable pre-PR on every live poll tick — not a regression
  here, but worth a follow-up sweep (add `Reverting` to both).
  - **Pre-existing/dormant: gRPC ordered-cutover preflight** — the adoption
  set at
  [grpc_client.go#L1764-L1777](https://github.com/block/schemabot/blob/9e93468
  4bc4556cbab877ad0609e68dfd23f530f/pkg/tern/grpc_client.go#L1764-L1777)
  handles `RevertWindow`/`SkippingRevert` but not `Reverting` (and the
  `SkippingRevert` arm is dead code — no proto state produces it), causing
  bounded ~1/min claim churn for a mid-revert remote until it terminalizes.
  Dormant until multi-deployment fan-out lands.
  - **Altitude** — the new invariant ("every dwellable non-terminal state must
  be claimable") is prose-only in a hand-maintained list; `pkg/state`'s
  metadata registry + completeness-test pattern could enforce it mechanically,
  preventing the next dwellable state from recreating this bug. Relatedly,
  the operation-side orphan claim inherits the fix via the shared
  `claimableApplyStates()` but has no revert-phase test.

  ### Verified correct

  - Both claim paths are genuinely fixed: `FindNextApply`, `ClaimApplyByID`,
  and `FindNextApplyOperation`'s stale-active clause all build from
  `claimableApplyStates()`
  ([applies.go#L114-L115](https://github.com/block/schemabot/blob/9e934684bc45
  56cbab877ad0609e68dfd23f530f/pkg/storage/mysqlstore/applies.go#L114-L115)),
  so the "parent apply not claimable for operation" loop resolves for these
  states.
  - The claim only rotates the lease; the persisted revert-phase state is
  preserved (asserted by the new `ClaimApplyByID` test, confirmed in the SQL).
  - `taskInRevertPhase` task-state coverage is complete: `skipping_revert`
  applies hold `revert_window` tasks (`complete_pending_revert` mapping),
  `reverting` maps to task `reverting`; no task-level skip state exists to
  miss.
  - A reverting apply can never hit the sequential resume path (Vitess always
  grouped, `local_apply.go:337`), which would have flipped it to `running` and
  re-run tasks.
  - `markApplyRecovering`'s task rewrite to `recovering` cannot blind the
  guard: its branch is gated to `waiting_for_cutover`/`recovering` applies
  only.
  - Test staleness offset (2 min) is 2× `ApplyLeaseStaleAfter`; the fixture
  engine is genuinely `planetscale` (derived from the vitess lock, not a
  default); the fresh-lease-not-stolen assertion is discriminating.
  - The raw `"recovering_cutover"` string in the claimable list is a
  deliberate legacy alias (normalized in `pkg/state/resolve.go`), not a
  missing constant.
  - `mysqlstore` is the sole store implementation — no parallel state list
  missed; `pkg/state/metadata.go` already classifies both states non-terminal.
  - CI fully green (all 33 checks, including E2E Vitess/K8s matrices).

aparajon:
🤖 Thanks for the deep review — both findings were real. All addressed in 7e9aeaca:

**Finding 1 (post-reattach persistence stomps revert-phase states) — confirmed and fixed.** The persistence step after the engine accepts the reattach now preserves revert-phase task and apply states instead of projecting everything to `running`, so the marker `taskInRevertPhase` keys on survives repeated driver deaths. I extracted the step into `persistReattachedResumeStates` so the projection is directly testable — `TestPersistReattachedResumeStates_PreservesRevertPhase` covers `revert_window` / `reverting` / `skipping_revert`, and `TestPersistReattachedResumeStates_ProjectsForwardStates` proves the normal forward projection (and the `recovering` carve-out) still holds.

**Finding 2 (missing resume state falls back to a fresh engine apply) — fixed fail-closed.** `groupedResumeState` now returns `errGroupedResumeStateUnavailable` when a revert-phase Vitess apply has no persisted engine resume state, instead of handing the engine a context-only state that takes the fresh-apply path and re-deploys the reviewed DDL on top of the revert. Covered by `TestGroupedResumeStateFailsClosedForRevertPhaseVitessApply`.

Also picked up the non-blocking notes:
- **Reverted tasks in the re-plan**: terminal `reverted` tasks now stay terminal instead of being re-activated. Taking that skip naively would have routed an all-reverted apply into the empty-active-tasks branch, which hardcoded `completed` — so that branch now derives the terminal state from the task states via `state.DeriveApplyState` (all-reverted terminalizes as `reverted`). `TestLaunchAtomicResumeAllRevertedTasksTerminalizeAsReverted` proves it end to end.
- **Duplicate `database` slog key** on the revert-phase Info log: dropped (already in `LogAttrs()`).
- **applies.go doc overclaim**: softened — a crash between the engine accepting the command and the durable write can re-issue it; the engine treats that as idempotent.
- **Discriminating re-plan test input**: the "revert already landed" subtest now plans a divergent column type so the completion branch is provably driven by the schema comparison rather than a vacuous diff.

The pre-existing items you flagged (cancel-mid-revert terminal sweep, gRPC ordered-cutover preflight, state-registry completeness test) are queued as follow-ups rather than folded in here.

*Generated by Claude Code (Claude Fable 5).*

Kiran01bm (approved):

PR #785, opened 2026-07-21
refactor(webhook): extract error-returning aggregate fold core

Kiran01bm:
## Summary

Extracts a one-shot aggregate-fold core that returns a typed follow-up disposition plus the underlying error, preparing the durable `check_run.completed` path to distinguish operational failures from pending participant convergence. Current fire-and-forget behavior is preserved.

## What

- Adds `updateAggregateCheckOnce(...) (aggregateFoldFollowUp, error)`.
- Keeps `updateAggregateCheck(...)` as a behavior-preserving wrapper that applies the returned follow-up and ignores the error (matching today's void semantics).
- Head-SHA verification now distinguishes a superseded head `(false, nil)` from an operational error `(false, err)`.
- Per-environment upserts attempt every configured environment and aggregate failures with `errors.Join`.
- Adds characterization tests for the follow-up/error contract.

## Why

The durable completed-event path must tell apart:

- successful fold, participant convergence pending → participant follow-up + nil error
- failed fold that should be retried → follow-up + non-nil error

The previous void-returning fold could not express that distinction. This slice makes the distinction available without changing runtime behavior yet.

## Before / after

```text
Before                              After
updateAggregateCheck (void)         updateAggregateCheck (wrapper)
 - folds aggregate                   - calls one-shot core
 - schedules follow-ups inline       - applies returned follow-up
                                     - preserves current behavior

                                    updateAggregateCheckOnce
                                     - returns follow-up + error
                                     - no timer/budget side effects
                                     - joins per-environment errors
```

## Testing / validation

- `go build ./...`
- `go vet ./pkg/webhook/`
- `go test -race ./pkg/webhook/ -count=1`
- `golangci-lint run --new-from-rev origin/main ./pkg/webhook/`

## Scope

The 1a aggregate-upsert error plumbing is now on `main` (#784). This PR contains only the behavior-preserving fold-core extraction. Durable `check_run.completed` dispatch wiring remains a separate WH-9d follow-up.

### DIFF
--- a/pkg/webhook/aggregate_fold_core_test.go
+++ b/pkg/webhook/aggregate_fold_core_test.go
@@ -59,7 +59,11 @@ func (s *foldStorage) Checks() storage.CheckStore {
 func newFoldHandler(t *testing.T, cfg *api.ServerConfig, checks storage.CheckStore) (*Handler, *http.ServeMux, *ghclient.InstallationClient) {
 	t.Helper()
 	client, mux := setupGitHubServer(t)
-	installClient := ghclient.NewInstallationClient(client, testLogger())
+	// Resolve the App slug so a participant Check Run read that finds no run
+	// classifies as a clean "not reported" (retriable, no error) rather than an
+	// ownership-unverifiable read error — letting a test exercise the genuine
+	// pending-convergence path distinctly from a participant read failure.
+	installClient := ghclient.NewInstallationClientWithSlug(client, testLogger(), "schemabot")
 	h := &Handler{
 		service:   api.New(&foldStorage{checks: checks}, cfg, nil, testLogger()),
 		ghClients: ghclient.NewSingleClientSet(defaultAppName, &fakeClientFactory{client: installClient}),
@@ -77,6 +81,32 @@ func serveHeadSHA(t *testing.T, mux *http.ServeMux, sha string) {
 			"head": map[string]any{"sha": sha, "ref": "feature-branch"},
 			"base": map[string]any{"sha": "def456", "ref": "main"},
 			"user": map[string]any{"login": "testuser"},
+		}))
+	})
+}
+
+// serveLeaderPRFiles registers the changed-files lookup the aggregate leader
+// performs to resolve which participants a PR requires, reporting the given path
+// as modified so the participant covering it becomes expected.
+func serveLeaderPRFiles(t *testing.T, mux *http.ServeMux, path string) {
+	t.Helper()
+	mux.HandleFunc("GET /repos/octocat/hello-world/pulls/1/files", func(w http.ResponseWriter, _ *http.Request) {
+		require.NoError(t, json.NewEncoder(w).Encode([]map[string]string{{
+			"filename": path,
+			"status":   "modified",
+		}}))
+	})
+}
+
+// serveParticipantCheckRunsEmpty registers the commit check-runs lookup
+// FindCheckRunByName performs, reporting no runs on the commit so an expected
+// participant reads as not-yet-reported (a retriable, still-pending outcome).
+func serveParticipantCheckRunsEmpty(t *testing.T, mux *http.ServeMux, sha string) {
+	t.Helper()
+	mux.HandleFunc("GET /repos/octocat/hello-world/commits/"+sha+"/check-runs", func(w http.ResponseWriter, _ *http.Request) {
+		require.NoError(t, json.NewEncoder(w).Encode(map[string]any{
+			"total_count": 0,
+			"check_runs":  []any{},
 		}))
 	})
 }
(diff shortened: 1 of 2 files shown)


aparajon:
🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head 568df66. I checked the wrapper's behavior preservation branch by branch against the old void fold and then attacked the new follow-up/error contract from the perspective of the durable `check_run.completed` path that will consume it. The extraction itself is clean and genuinely behavior-preserving; the two items below are about the contract, not the refactor.

### Findings / questions

**1. Participant Checks-API read failures don't surface in the error return, which undercuts the stated contract (main question)**

The doc comment on `aggregateFoldFollowUp` says a `ScheduleParticipantRefold` follow-up with a **nil** error is "a successful fold whose participant convergence is still pending — distinct from a failed fold, which also returns a non-nil error." But `resolveParticipantOutcome` (`pkg/webhook/leader_participant_gate.go`) maps a `FindCheckRunByName` error to a blocked+retriable outcome and never propagates the error, so a fold in which **every participant read failed operationally** returns `(aggregateFoldScheduleParticipantRefold, nil)` — indistinguishable from genuine pending convergence. The asymmetry is stark on the leader path specifically: this PR makes the leader's PR-files fetch failure return an error, while the participant check-run read failure — an equally operational GitHub read two lines later in the same fold — returns nil.

This matters for the consumer this PR is preparing: with a nil error the durable event presumably completes, and convergence for that failure class then rests solely on the participant-refold timer chain — which is in-memory (`participantRefoldAttempts` + `time.AfterFunc`) and doesn't survive a pod restart. If the durable path is meant to be the cross-restart safety net for operational failures, this class is invisible to it.

I can see the argument for the current shape (folding these errors in would make the durable retry and the refold timer double-retry the same condition, and the refold already owns participant convergence). If that's the intent, I'd suggest saying so explicitly in the `aggregateFoldFollowUp` doc comment — "participant *read* failures are deliberately conveyed only through the retriable disposition, not the error" — because as written the comment overclaims and the next person wiring the durable path will reasonably assume nil error ⇒ no operational failure occurred.

**2. The headline disposition — `(ScheduleParticipantRefold, nil)` — is the one case with no test**

The characterization suite covers nine branches, but every `ScheduleParticipantRefold` case it pins carries a non-nil error (leader can't fetch PR files). The "successful fold, participant convergence pending → participant follow-up + nil error" case — the exact distinction the PR body says the durable path needs — is untested. A leader config with an expected participant whose check-run lookup returns no run (or errors, per finding 1) plus a successful upsert would pin it, and would also lock in whichever answer you choose for finding 1.

### Nits

- The wrapper's `switch followUp` has no `default:`. Go won't flag a new `aggregateFoldFollowUp` constant added later, so it would silently no-op in the fire-and-forget path — a `default:` that logs an error would keep the "no silent branch cases" property when the enum grows.

### Verified (tried to break, couldn't)

- **Behavior preservation, branch by branch.** All six old exit paths map to identical side effects through the wrapper: checks-disabled → nothing; empty head / PR-fetch failure / superseded head → `scheduleLeaderRefoldIfConfigured`; `GetByPR` failure → `scheduleLeaderRefoldIfConfigured`; leader PR-files failure → `scheduleParticipantRefold`; no checks and no participants → nothing; retriable participants → `scheduleParticipantRefold`, else `clearParticipantRefoldBudget`. The upsert loop's runtime behavior is unchanged too — the old fire-and-forget wrapper already discarded the same errors the core now collects.
- **No caller drift.** All 15 production call sites remain on the void `updateAggregateCheck` wrapper; the deleted `upsertAggregateCheckRun` wrapper has no remaining callers; `verifyHeadSHAStillCurrentForPR` keeps byte-identical bool semantics, metrics, and logs for its other callers, and `(true, non-nil)` is unrepresentable by construction.
- **Error classification at the head-SHA gate is right.** A genuinely superseded head returns `(false, nil)` — correctly *not* retriable as an operational failure (retrying a stale-head fold as if GitHub had failed would be wrong), while empty-head and PR-fetch failures return errors. This is exactly the classification the durable path needs at that gate.
- **`errors.Join` usage is sound.** Only non-nil errors are appended, and `errors.Join()` of an empty slice is nil, so the success path returns a nil error; the per-environment loop attempts every environment after an earlier failure and the test pins both the attempt count and the joined error.
- **Fail-closed posture preserved.** Budget exhaustion still leaves the aggregate blocked with a warn + exhausted metric; no path converts a storage or GitHub uncertainty into a passing aggregate.

aparajon (approved):

Kiran01bm:
> 🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head [568df66](https://github.com/block/schemabot/commit/568df66f63ca75e28045e89fda837da36a249e48). I checked the wrapper's behavior preservation branch by branch against the old void fold and then attacked the new follow-up/error contract from the perspective of the durable `check_run.completed` path that will consume it. The extraction itself is clean and genuinely behavior-preserving; the two items below are about the contract, not the refactor.

| # | Concern | Status | Explanation |
|---|---------|--------|-------------|
| 1 | Participant Check Run read failures return `(ScheduleParticipantRefold, nil)`, so the doc comment's "nil error ⇒ no operational failure" overclaims | **fixed** | Confirmed. Took the reviewer's stated-intent path: rewrote the `aggregateFoldFollowUp` doc comment to say participant *read* failures are **deliberately** conveyed only through the retriable disposition, not the error return — folding them in would make a retry-owning caller and the re-fold timer double-retry the same condition. The comment now states a nil error asserts only that the fold completed and the aggregate was written, not that no participant read failed. |
| 2 | The headline `(ScheduleParticipantRefold, nil)` disposition — successful fold, participant convergence pending — is the one case with no test | **fixed** | Added two characterization cases: a genuine *not-yet-reported* participant (clean missing read) and a participant *read failure*, both pinning `(ScheduleParticipantRefold, nil)`. The two now exercise distinct branches (verified via logs: "has not reported" vs "failed to query participant check"), locking in the finding-1 decision. |
| 3 | Nit: wrapper `switch followUp` has no `default:`, so a new disposition would silently no-op | **fixed** | Added a `default:` that logs an error ("unhandled aggregate fold follow-up, post-fold side effect dropped") so a future enum value can't drop its side effect unnoticed — preserves the repo's "no silent branch cases" property. |

PR #788, opened 2026-07-21
fix(observability): emit per-table ETA for non-sharded tables

aparajon:
## Why this matters

Operators watching a long copy on a non-sharded (plain MySQL) table saw no ETA in the CLI progress view or PR progress comments, even though the data plane computes and stores one on every progress poll. Sharded tables already surfaced an ETA through the per-shard aggregate, so the two engines gave inconsistent progress UX.

## What it does

- Copies the stored task's ETA into the proto `TableProgress` in the non-sharded path of the progress assembly, mirroring how the other stored progress figures (rows, percent, checksum) are already copied. The per-shard aggregate path is unchanged.
- Adds a focused test proving a non-sharded table's headline figures — including the ETA — render from the stored task row.

---

### DIFF
--- a/pkg/api/progress_handlers_test.go
+++ b/pkg/api/progress_handlers_test.go
@@ -1,0 +1,50 @@
+package api
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/block/schemabot/pkg/apitypes"
+	ternv1 "github.com/block/schemabot/pkg/proto/ternv1"
+	"github.com/block/schemabot/pkg/state"
+)
+
+// A table's ETA travels from the progress proto through the HTTP response and
+// its JSON encoding unchanged, so CLI and API consumers see the same figure
+// the driver stored.
+func TestProgressResponseFromProtoCarriesTableETA(t *testing.T) {
+	resp := progressResponseFromProto(&ternv1.ProgressResponse{
+		State:  ternv1.State_STATE_RUNNING,
+		Engine: ternv1.Engine_ENGINE_SPIRIT,
+		Tables: []*ternv1.TableProgress{
+			{
+				TableName:       "users",
+				Namespace:       "testdb",
+				Status:          "running",
+				RowsCopied:      250000,
+				RowsTotal:       500000,
+				PercentComplete: 50,
+				EtaSeconds:      540,
+			},
+		},
+	})
+
+	require.Len(t, resp.Tables, 1)
+	assert.Equal(t, int64(540), resp.Tables[0].ETASeconds)
+
+	encoded, err := json.Marshal(resp)
+	require.NoError(t, err)
+	var decoded apitypes.ProgressResponse
+	require.NoError(t, json.Unmarshal(encoded, &decoded))
+
+	require.Len(t, decoded.Tables, 1)
+	assert.Equal(t, "users", decoded.Tables[0].TableName)
+	assert.Equal(t, state.Apply.Running, decoded.State)
+	assert.Equal(t, int64(250000), decoded.Tables[0].RowsCopied)
+	assert.Equal(t, int64(500000), decoded.Tables[0].RowsTotal)
+	assert.Equal(t, int32(50), decoded.Tables[0].PercentComplete)
+	assert.Equal(t, int64(540), decoded.Tables[0].ETASeconds)
+}
--- a/pkg/cmd/commands/apply.go
+++ b/pkg/cmd/commands/apply.go
@@ -856,6 +856,10 @@ func recoveringLogMessage(tbl *apitypes.TableProgressResponse) string {
 func (e *logEmitter) emitProgressHeartbeat(tbl *apitypes.TableProgressResponse, ts *tableLogState) {
 	kvs := tableKVs("Copying rows", tbl, ts)
 	if ui.EstimateExceeded(tbl.RowsCopied, tbl.RowsTotal) {
+		// A table past its row estimate has outlived the figure the ETA was
+		// derived from — in a multi-table apply the shared runner estimate can
+		// still be nonzero here, so appending it would contradict the
+		// "Finalizing copy" wording.
 		kvs = append(kvs,
 			"progress", "Finalizing copy",
 			"rows_copied", fmt.Sprintf("%s so far", ui.FormatNumber(tbl.RowsCopied)),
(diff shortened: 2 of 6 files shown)


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/788.

**Verdict:** clean — approve. The one-line fix is correct, loses nothing, and does exactly what the PR claims; every candidate bug the review surfaced turned out to be either pre-existing parity or unrendered residue. CI is green across the board, including the E2E MySQL suite that exercises the changed path.

The risk in surfacing a previously-dead field is stale data: `ETASeconds` is copied from the stored task row [unconditionally](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/tern/local_client.go#L2291), with no terminal-state or freshness guard — a frozen pre-crash ETA, a failed task's leftover estimate, or a completed task terminalized without a final poll could all now render. I chased each of those and none is a regression this PR introduces:

- **Failed/stopped/completed tasks:** every renderer status-gates ETA out of terminal states — the CLI prints bar+label only for Failed/FailedRetryable, the webhook comment explicitly renders "rows (no ETA) for stopped tables", and neither prints ETA for Completed. The only residual exposure is the raw `eta_seconds` JSON field on `GET /progress` ([progress_handlers.go:183](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/api/progress_handlers.go#L183)), and no in-repo consumer keys logic off `eta > 0`.
- **Recovering/stalled-running windows** do render a frozen ETA — but the PR-comment surface already did exactly this before the PR, because [`tableProgressFromTasks`](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/webhook/apply.go#L115) builds comment data straight from the storage rows (`ETASeconds: int64(t.ETASeconds)`), bypassing `Progress()` entirely. The sharded path in the same function has identical unguarded semantics ([`aggregateStoredShards`](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/tern/local_client.go#L2524) even maxes over completed shards' ETAs). So the PR brings the non-sharded ternv1/CLI path to parity with two pre-existing surfaces rather than losing a guard either of them had — and the window self-heals on the first engine poll after reattach, since Spirit reports 0 until its estimate is `ETAReady`.
- **Mechanics:** `storage.Task.ETASeconds` is `int`, so `int64()` is a lossless widening; Spirit zeroes ETA for completed tables and past 99.99% copy; for Vitess the new line only fills the window before shard rows are persisted (the per-shard aggregate [overwrites it at 2308-2312](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/tern/local_client.go#L2308) once they exist — an improvement over the previous blank, and the −1 "unknown" sentinel is filtered by max-with-0 on both sides).

### Notes (not blocking)

- **The per-table field carries Spirit's runner-wide estimate.** [`buildSpiritTableProgress`](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/engine/spirit/spirit.go#L561) documents that Spirit reports one remaining-copy estimate for the whole runner and stamps it on every still-copying table. In a multi-table apply, a nearly-done table can therefore display the ETA of the whole run. This is deliberate, documented, and already what the PR comment showed — but now that the value is per-table on CLI/API too, a future per-table estimate (or a doc note on the proto field) would remove the ambiguity.
- **No terminal-transition cleanup of `ETASeconds`.** `markTaskFailed`/`markTaskRetryable` and the resume re-plan completion paths ([local_control_resume.go:335](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/tern/local_control_resume.go#L335), [:475](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/tern/local_control_resume.go#L475)) leave the last ETA on terminal rows (the latter set `ProgressPercent=100` but not `ETASeconds=0` — the same asymmetry as the adjacent completed-clamp, which fixes percent/rows but not ETA). Renderers hide it today; zeroing ETA on terminal transitions would make the stored rows self-consistent. Applies equally to the pre-existing sharded path.
- **Coverage stops at the proto.** The new test pins `LocalClient.Progress`; downstream, no test feeds a nonzero `EtaSeconds` through `progress_handlers.go` → apitypes JSON → `progress_parse.go` → renderer (those tests hand-set fixtures). A future refactor could re-zero the field on that path with every test still green.
- **Multi-table estimate-exceeded edge:** the heartbeat log appends `eta` even in the "Finalizing copy … so far" branch ([apply.go:859-873](https://github.com/block/schemabot/blob/8edb1c68a36e3164068fb256a8df7d8241c823a4/pkg/cmd/commands/apply.go#L859)). Single-table this can't fire (Spirit zeroes ETA past 99.99%), but in a multi-table apply one table can exceed its own estimate while the aggregate is below the cutoff, emitting an `eta` derived from the shared runner estimate alongside the "estimate exceeded" wording. Cosmetic log-only inconsistency.

### Verified correct

- The copy sits with the other stored-field copies and runs before the per-shard aggregate, so ordering is right for sharded tables and the fallback fills a previously-blank window.
- Type conversion is lossless (`int` → `int64`); no sentinel value can surface (Spirit gates on `ETAReady`; the Vitess −1 is filtered by max-with-0 in both the engine and the aggregate).
- All render surfaces gate ETA by state — completed/failed/stopped tables show no ETA anywhere.
- The test uses the pre-existing `exactProgressStorage` fixture family (used by ~10 sibling tests), asserts all headline figures including the empty shard list, and passes; CI is fully green including E2E MySQL, Vitess, and integration suites.
- Reuse/simplification/efficiency and repo-conventions angles all came back clean.

aparajon:
🤖 Thanks for the thorough pass — all four notes addressed in 1fda0df7:

- **Runner-wide estimate:** documented at the `EtaSeconds` copy site in `LocalClient.Progress` that Spirit's stored figure is the runner-wide remaining-copy estimate stamped on every still-copying table.
- **Terminal-transition cleanup:** `transitionTaskState` now zeroes `ETASeconds` whenever a task leaves an in-flight state — this covers `markTaskFailed`/`markTaskRetryable` and both resume re-plan completion paths in one place, since they all funnel through it. A resumed or retried task gets a fresh figure from the first engine poll.
- **Pipeline coverage:** added `TestProgressResponseFromProtoCarriesTableETA` (proto → handler → apitypes JSON) and `TestParseProgressResponseCarriesTableAndShardETA` (JSON → render data, incl. per-shard ETA), so a future refactor that re-zeroes the field on that path fails a test.
- **Estimate-exceeded heartbeat:** the CLI heartbeat now only appends `eta` alongside normal copy progress — the "Finalizing copy" branch no longer emits one.

*This reply was posted by Armand's agent (Claude Code, claude-fable-5).*

Kiran01bm (approved):

PR #789, opened 2026-07-21
fix(github): minimize stale plan comments when auto-plan finds no changes

aparajon:
## Why this matters

Plan comments were only minimized when a newer plan comment posted. An auto-plan that resolves to no changes posts nothing (by design, to reduce PR noise) — so when a later commit makes the schema up to date, the previous head's plan comment stayed expanded, advertising pending DDL and a live apply prompt that no longer match the branch. Someone had to hide it manually. Staleness is a property of the head moving on, not of a new comment posting: an up-to-date outcome supersedes prior plan comments just as much as a new plan does.

## What it does

- The no-changes auto-plan outcome now runs the supersede-minimize sweep for the database slot before returning: prior unminimized plan comments rendered at other heads collapse, without posting any new comment.
- Before minimizing anything, the sweep verifies the delivery's head is still the PR's current head with a fresh (uncached) PR fetch. A sweep computed for an older head skips entirely — otherwise a slow no-changes plan racing a newer push could hide the newer head's live comment with nothing replacing it, and minimized comments are never automatically restored.
- Extracts the sweep into `minimizePlanCommentsForSlot`, shared by the posted-comment path (unchanged supersession via `planCommentSupersedes`) and the new head-based `minimizeStalePlanComments` variant.
- Safety semantics are unchanged: apply-owned comments stay expanded as the operational record, same-head comments stay expanded (they may be the only visible plan for their environment scope), and every failure keeps the comment expanded for retry on the next sweep.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -814,6 +814,11 @@ for the same set of environments. A same-commit comment covering different
 environments is kept expanded — it may be the only visible plan for those
 environments.
 
+A plan outcome can also supersede without posting: when an auto-plan resolves
+to no changes, no new comment appears (the check run alone reports the green
+state), but plan comments from prior commits still collapse — the pending DDL
+and apply prompt they show no longer match the branch.
+
 One safety hold: a plan comment whose commit produced an apply is never
 minimized, even after new pushes. That comment is the record of what actually
 ran against the database, and it stays visible until an operator reconciles
--- a/pkg/webhook/auto_plan_integration_test.go
+++ b/pkg/webhook/auto_plan_integration_test.go
@@ -462,6 +462,12 @@ func TestE2EAutoPlanWithLintViolations(t *testing.T) {
 	}
 }
 
+// TestE2EAutoPlanNoChangesSkipsComment exercises the up-to-date-PR UX through
+// the full webhook path: a synchronize delivery whose auto-plan resolves to no
+// changes posts no plan comment (the check run alone reports the green state),
+// and the outcome supersedes the slot's plan comments from prior heads — their
+// pending DDL and apply prompt no longer match the branch, so they collapse
+// even though no new comment replaces them.
 func TestE2EAutoPlanNoChangesSkipsComment(t *testing.T) {
 	dbName := "webhook_autoplan_nochange"
 	svc := setupE2EService(t, dbName)
@@ -489,6 +495,26 @@ func TestE2EAutoPlanNoChangesSkipsComment(t *testing.T) {
 
 	result := setupFakeGitHubForPlan(t, mux, schemaFiles, schemabotConfig, dbName)
 
+	// A plan comment from a prior head is still expanded on the PR; the
+	// no-changes outcome must collapse it.
+	insertPlanCommentRow(t, svc.Storage(), "octocat/hello-world", 1, dbName, "staging", "priorsha", 9001, "IC_stale_prior")
+
+	minimized := make(chan string, 4)
+	mux.HandleFunc("POST /graphql", func(w http.ResponseWriter, r *http.Request) {
+		var gql struct {
+			Variables map[string]string `json:"variables"`
+		}
+		assert.NoError(t, json.NewDecoder(r.Body).Decode(&gql))
+		minimized <- gql.Variables["id"]
+		_ = json.NewEncoder(w).Encode(map[string]any{
+			"data": map[string]any{
+				"minimizeComment": map[string]any{
+					"minimizedComment": map[string]any{"isMinimized": true},
+				},
+			},
+		})
+	})
+
 	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelError}))
 	installClient := ghclient.NewInstallationClient(client, logger)
 	factory := &fakeClientFactory{client: installClient}
@@ -515,6 +541,18 @@ func TestE2EAutoPlanNoChangesSkipsComment(t *testing.T) {
 	case <-time.After(30 * time.Second):
 		t.Fatal("timed out waiting for check run")
 	}
+
+	// The prior head's plan comment collapses even though no new comment is
+	// posted: the no-changes outcome supersedes it.
+	select {
+	case node := <-minimized:
+		assert.Equal(t, "IC_stale_prior", node)
+	case <-time.After(30 * time.Second):
+		t.Fatal("timed out waiting for the prior head's plan comment to be minimized")
+	}
+	require.Eventually(t, func() bool {
+		return len(unminimizedHeads(t, svc.Storage(), "octocat/hello-world", 1, dbName, "mysql")) == 0
+	}, 10*time.Second, 100*time.Millisecond, "the minimized plan comment must be recorded in storage")
 
 	// No comment should be posted — give it a moment to confirm nothing arrives
 	select {
--- a/pkg/webhook/plan.go
+++ b/pkg/webhook/plan.go
@@ -458,7 +458,14 @@ func (h *Handler) handleMultiEnvPlan(repo string, pr int, databaseName, tenant s
 			}
 		}
 		if !anyChanges && !hasErrors {
-			h.logger.Info("auto-plan: no changes detected, skipping comment", "repo", repo, "pr", pr)
+			// The no-changes outcome supersedes older plan comments just as a
+			// new plan comment would: a prior head's comment still advertises
+			// pending DDL and an apply prompt that no longer match the branch.
+			h.logger.Info("auto-plan: no changes detected; skipping comment and minimizing plan comments from prior heads",
+				"repo", repo, "pr", pr, "database", multiEnvData.Database,
+				"database_type", multiEnvData.DatabaseType, "head_sha", multiEnvData.HeadSHA)
+			h.minimizeStalePlanComments(ctx, client, repo, pr,
+				multiEnvData.Database, multiEnvData.DatabaseType, multiEnvData.HeadSHA)
 			return
 		}
 	}
--- a/pkg/webhook/plan_comment_minimize.go
+++ b/pkg/webhook/plan_comment_minimize.go
@@ -90,28 +90,58 @@ func (h *Handler) postTrackedPlanComment(repo string, pr int, installationID int
 }
 
 // minimizeSupersededPlanComments collapses the unminimized plan comments that
-// the newly posted comment supersedes. A comment whose plan became an apply
-// is kept expanded: the apply makes it the operational record of what ran,
-// and hiding it costs more than the noise it saves. Every failure keeps the
-// comment expanded and its row unminimized, so the next supersede retries it.
+// the newly posted comment supersedes, per planCommentSupersedes.
 func (h *Handler) minimizeSupersededPlanComments(ctx context.Context, client *ghclient.InstallationClient, posted *storage.PlanComment) {
+	h.minimizePlanCommentsForSlot(ctx, client,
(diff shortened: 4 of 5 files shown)


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/789.

**Verdict:** 2 findings — a stale-head race can permanently hide the PR's live plan comment (the new sweep hides without replacing, unlike the pre-existing posted-path race), and the PR's only production wiring change is unpinned by tests (deleting the call leaves the whole suite green).

### Findings

**1. CONFIRMED (race, low frequency / high impact) — a slow no-changes auto-plan for an older head minimizes the newer head's live plan comment, and nothing ever restores it.** [plan_comment_minimize.go:158](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize.go#L158)

Every link verified: auto-plans run as detached per-config goroutines (`h.goSafe`, [pull_request.go:356-358](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/pull_request.go#L356-L358)) with no per-PR serialization anywhere — the durable inbox claim is a global-FIFO `SKIP LOCKED` query with no per-PR predicate ([webhook_events.go:193-200](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/storage/mysqlstore/webhook_events.go#L193-L200)), and [durable_dispatch.go:423-428](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/durable_dispatch.go#L423-L428) documents that plan execution stays detached. The sweep's head is the goroutine's delivery-cached `FetchPullRequest` value (up to the 2-minute command window stale), and the freshness gate is explicitly skipped for auto-plan (`if !isAutoPlan {`, [plan.go:359](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan.go#L359)) — justified by the comment at [plan.go:353-358](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan.go#L353-L358) ("covered by the next synchronize delivery superseding any stale comment"), which is precisely the assumption this sweep breaks: it *hides* the newer delivery's comment instead of being superseded by it. With `posted == nil` the sweep minimizes every slot comment at any *different* head — newer included ([plan_comment_minimize.go:158](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize.go#L158)) — and no un-minimize path exists anywhere (`minimizeComment` is the only GraphQL mutation; nothing clears `minimized_at`). Concrete scenario: push A plans to no changes slowly; push B (re-adding a change) lands inside that window and its goroutine posts the live comment at shaB; the shaA goroutine then reaches the sweep and hides shaB's comment — the PR ends with pending DDL, a live apply prompt, and **zero visible plan comments** until the next push or a manual `schemabot plan`. The direction-blind supersede pre-exists on the posted path ([planCommentSupersedes](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize.go#L203-L208)), but that race leaves a stale comment as replacement; this one leaves nothing. Frequency is genuinely narrow (two pushes within the older goroutine's window, older head fetched before the newer push, older goroutine finishing last) — but the trigger population, PRs churning through convergence/reverts, is exactly who this PR serves. Cheap guard already in-repo: a `FetchPullRequestNoCache` head check before sweeping (the manual-plan gate at [plan.go:360](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan.go#L360) uses it; note `verifyHeadSHAStillCurrentForPR` would NOT work here — it reads the same cached fetch).

**2. CONFIRMED (test fidelity, empirically proven) — the wiring from the no-changes branch to the sweep is the PR's only production behavior change, and no test exercises it.** [plan.go:467](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan.go#L467)

A finder deleted the `h.minimizeStalePlanComments(...)` call from plan.go and the full suite passed (run twice, worktree restored). Both new tests call the helper directly ([plan_comment_minimize_integration_test.go:372](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize_integration_test.go#L372), [:433](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize_integration_test.go#L433)), so they pin the helper's semantics but not the call. The existing end-to-end no-changes test ([auto_plan_integration_test.go:465](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/auto_plan_integration_test.go#L465)) drives the exact branch but seeds no `plan_comments` rows and its fake mux registers no `/graphql` handler, so the sweep is an invisible no-op there. A future restructuring of `handleMultiEnvPlan`'s `isAutoPlan` block can silently drop the feature with CI green. Cheap fix: seed a prior-head row via the PR's own `insertPlanCommentRow` helper in that e2e test (plus the graphql minimize handler from the minimize test file) and assert the stale node collapsed.

The refactor folds the load-bearing posted-comment supersede path into a shared helper — a subtle behavior change there would corrupt plan-comment lifecycle on every PR. Verified byte-equivalent: the wrapper passes `posted.Repository, posted.PullRequest, posted.DatabaseName, posted.DatabaseType, posted.HeadSHA` into the parameter list in exactly that order (no transposition; the slot query takes the same argument set as base), and for `posted != nil` the self-skip and `planCommentSupersedes` logic are untouched — a diff against `origin/main` shows only log-wording and parameterization changes. The new nil mode is strictly more conservative at the same head (keeps *all* same-head comments, where the posted path minimizes same-head refreshes), the apply-owned hold sits on the shared path after both supersede branches with correct bookkeeping (the hold `continue`s before `MarkMinimized`, so held rows stay live for future evaluation), slot scoping is airtight (`WHERE repository/pull_request/database_name/database_type`, and `MinimizeComment` uses the prior row's own repo/node), and concurrent or repeated sweeps are idempotent (GitHub's minimize mutation succeeds on an already-minimized node; `MarkMinimized` is guarded by `minimized_at IS NULL`). The nil-posted design is also semantically necessary, not API bloat — a synthetic `posted` record would flow through `planCommentSupersedes` and wrongly minimize the current head's own comments while falsely matching zero-ID rows in the self-skip.

### Notes (not blocking)

- **docs/configuration.md now understates the behavior:** the "Plan comment minimization" section ([docs/configuration.md:803](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/docs/configuration.md#L803)) describes minimization solely as "the plan comments a newer one supersedes" — an operator whose comment collapses with no new comment posted will find the observed behavior undocumented. One sentence fixes it.
- **The most common convergence shape never reaches the sweep:** a push that *removes* the schema change from the PR entirely (or drops one database) means config discovery finds nothing for that slot and `handleMultiEnvPlan` never runs — the prior head's pending-DDL comment stays expanded forever, exactly the staleness the PR body says it fixes. `cleanupStaleChecks` ([pull_request.go:621](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/pull_request.go#L621)) already handles removed databases for check records and would be the natural sweep site for a follow-up.
- **Cadence-skip deliveries sweep too:** the sweep precedes the `if !postPlanComment` return ([plan.go:473](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan.go#L473)), so a docs-only push that resolves to no changes still minimizes prior heads. Consistent with the PR body's rationale ("staleness is a property of the head moving on"), but undocumented, untested for that combination — and it widens finding 1's trigger surface.
- **The apply-owned test can't distinguish the hold from the error fallback:** `assert.Empty(fake.minimizedNodes())` also holds if `ExistsForDatabaseHead` errors (the `guard_error` path keeps comments expanded too). Adding a second, non-apply-owned stale row and asserting it *is* minimized would prove the sweep ran and isolate the hold. Matches the pre-existing sibling test's shape, so style-consistent.
- **Smaller coverage gaps:** the empty-identity guard ([plan_comment_minimize.go:107-114](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize.go#L107-L114)) is untested (its sibling in `postTrackedPlanComment` has one), and the minimize-failure retry is covered only through the posted path — low risk since the code is shared.
- **`minimizeSupersededPlanComments` is now a pure single-caller delegation** ([plan_comment_minimize.go:94-97](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/webhook/plan_comment_minimize.go#L94-L97)); inlining it at the post site would satisfy AGENTS.md's no-wrapper preference, though keeping the name for godoc symmetry is defensible.

### Verified correct

- `multiEnvData.Database/DatabaseType/HeadSHA` are guaranteed populated on the no-changes branch (every env-loop failure sets `Errors`, so `!hasErrors` implies a successful first iteration set all three from the live fetch) — the empty-identity guard is pure defense-in-depth.
- The branch is auto-plan-only (`if isAutoPlan`); manual plans always post (including no-changes) and sweep through the posted path, so manual staleness was already handled.
- The apply-owned hold's PR-scoping (`WHERE a.repository = ? AND a.pull_request = ?`, [applies.go:2083-2085](https://github.com/block/schemabot/blob/9cfe712b6b96cdb2389134c7deaef66986ab076a/pkg/storage/mysqlstore/applies.go#L2083-L2085)) is by design, not a defect — a cross-PR-applied change has its operational record on that PR's own comment.
- Cross-database/type/PR isolation asserted and proven by the new test's exact-match assertions (billing slot untouched, `createCount()==0` is meaningful — the fake counts every comment create).
- Error outcomes (`hasErrors`) intentionally don't sweep — an error outcome doesn't supersede a valid plan; and the with-changes cadence-skip path correctly leaves the prior comment as the still-accurate plan.
- AGENTS.md conventions hold: `fix(github)` scope correct for PR-comment UX, godocs describe invariants not history, every skip branch logs, `t.Context()` throughout; `go build`, `go vet` (with `-tags=integration`) clean; all 8 plan-comment-minimize tests pass under testcontainers; CI 32/32 green at the head.

aparajon:
🤖 Both findings addressed in 54d62bd8:

**1 (stale-head race):** `minimizeStalePlanComments` now verifies the sweep's head against a fresh `FetchPullRequestNoCache` before minimizing, exactly the guard you suggested. Head moved or fetch failed → the sweep skips and every comment stays expanded; the current head's own plan outcome supersedes the slot. Pinned by `TestStalePlanCommentSweepSkipsWhenHeadMoved`.

**2 (unpinned wiring):** the e2e no-changes test now seeds a prior-head plan-comment row and asserts it collapses through the webhook path. Re-ran your mutation check: deleting the `h.minimizeStalePlanComments` call now fails the test.

Also took the docs sentence and the apply-owned-test discrimination from the notes. The removed-database sweep (`cleanupStaleChecks`) is a good follow-up — left out of this PR.

*— Claude Code (Claude Fable 5)*

Kiran01bm (approved):

PR #790, opened 2026-07-21
feat(github): make check_run.rerequested durable

Kiran01bm:
## Summary

Moves the `check_run.rerequested` auto-plan off the request goroutine and onto the durable webhook inbox, gated behind the existing durable-dispatch flag. A deploy/SIGTERM mid-run previously dropped the re-plan; now a fast ACK persists a leased-driver row with heartbeats, capped retries, and lease-expiry reclaim — full parity with the `pull_request` path. When the flag is off, the synchronous path is unchanged.

## What

- HTTP `handleCheckRunRerequest` enqueues a durable row (`enqueueDurableCheckRun`) and acks when durable dispatch is enabled.
- Driver routes `check_run` → `processDurableCheckRunRerequest`: re-validates guards, resolves the installation ID, verifies the head is still current, then runs auto-plan.
- Head verification is retry-aware: a GitHub failure keeps the delivery retryable (never silently dropped); a confirmed newer head completes without planning a stale snapshot.
- Adds `rerequested` to the webhook metrics action allowlist so it is no longer folded to `unknown`.

## Why

`check_run.rerequested` was the last synchronous auto-plan entry point with an ack-then-drop window on deploy. Making it durable closes that window. The change is additive and independent of the fold-core refactor stack (which covers the separate `check_run.completed` path).

## Before / after
```
Before                                  After (flag on)
handleCheckRunRerequest                 handleCheckRunRerequest
runAutoPlanForPR synchronously          enqueue row + fast ack
(deploy mid-run drops it)              leased driver:
verify head (retry on GH error,
complete on stale)
runAutoPlanForPR
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -874,6 +874,7 @@ var knownWebhookActions = map[string]bool{
 	"ready_for_review":       true,
 	"reopened":               true,
 	"requested":              true,
+	"rerequested":            true,
 	"review_request_removed": true,
 	"review_requested":       true,
 	"submitted":              true,
--- a/pkg/webhook/check_run.go
+++ b/pkg/webhook/check_run.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+
+	"github.com/block/schemabot/pkg/metrics"
 )
 
 type checkRunPayload struct {
@@ -26,7 +28,7 @@ type checkRunPayload struct {
 	} `json:"installation"`
 }
 
-func (h *Handler) handleCheckRun(ctx context.Context, w http.ResponseWriter, body []byte) {
+func (h *Handler) handleCheckRun(ctx context.Context, metricApp string, w http.ResponseWriter, body []byte, deliveryID string) {
 	var payload checkRunPayload
 	if err := json.Unmarshal(body, &payload); err != nil {
 		h.writeError(w, http.StatusBadRequest, "invalid check_run payload")
@@ -35,7 +37,7 @@ func (h *Handler) handleCheckRun(ctx context.Context, w http.ResponseWriter, bod
 
 	switch payload.Action {
 	case "rerequested":
-		h.handleCheckRunRerequest(ctx, w, payload)
+		h.handleCheckRunRerequest(ctx, metricApp, w, payload, body, deliveryID)
 	case "completed":
 		h.handleParticipantCheckCompleted(ctx, w, payload)
 	default:
@@ -121,7 +123,7 @@ func (h *Handler) handleParticipantCheckCompleted(ctx context.Context, w http.Re
 	h.writeJSON(w, http.StatusOK, map[string]string{"message": "aggregate re-folded on participant check completion"})
 }
 
-func (h *Handler) handleCheckRunRerequest(ctx context.Context, w http.ResponseWriter, payload checkRunPayload) {
+func (h *Handler) handleCheckRunRerequest(ctx context.Context, metricApp string, w http.ResponseWriter, payload checkRunPayload, body []byte, deliveryID string) {
 	installationID := h.effectiveInstallationID(ctx, payload.Installation.ID)
 	if installationID == 0 {
 		h.writeError(w, http.StatusBadRequest, "missing installation ID in webhook payload")
@@ -135,7 +137,8 @@ func (h *Handler) handleCheckRunRerequest(ctx context.Context, w http.ResponseWr
 			"repo", repo,
 			"check_run_id", payload.CheckRun.ID,
 			"check_name", payload.CheckRun.Name,
-			"head_sha", payload.CheckRun.HeadSHA)
+			"head_sha", payload.CheckRun.HeadSHA,
+			"delivery_id", deliveryID)
 		h.writeJSON(w, http.StatusOK, map[string]string{"message": "check_run rerequest ignored without pull request"})
 		return
 	}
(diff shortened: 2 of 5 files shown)


aparajon:
🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head 9bde161, with the same lens as the #780 review (enqueue/drive guard symmetry, retry-vs-terminal classification, claim-path reachability). Verdict up front: this is a solid slice — the guard symmetry is complete, the retry classification is right everywhere I attacked it, and it even fixes the malformed-row-ordering asymmetry class I flagged on #780's push processor. One real finding, plus small notes.

### Findings

**1. The shutdown-detached enqueue is the right hardening — but it's asymmetric with `enqueueDurablePullRequest`, and the comment's justification implies a distinction that doesn't exist**

`enqueueDurableCheckRun` detaches the INSERT from request cancellation (`context.WithoutCancel` + 5s bound) so a graceful shutdown can't abort the persist and 500 the delivery. Good. But `enqueueDurablePullRequest` still runs its INSERT on the raw request context, so the exact scenario the new comment describes — enqueue aborted at the durability boundary, delivery lost until a manual Redeliver — remains open on the pull_request path. The comment's rationale ("GitHub never auto-retries a check_run rerequest") reads as if pull_request deliveries fare better, but GitHub doesn't auto-retry *any* failed webhook delivery; the pull_request path's only extra backstop is the report-only missing-delivery scan, which detects but doesn't redeliver. The hardening applies equally to both producers — I'd lift the detach into a shared enqueue helper (or apply it to `enqueueDurablePullRequest` in a follow-up) and state the real rationale rather than the event-type contrast.

### Notes (minor)

- The driver's installation-ID fallback (`effectiveInstallationID` on unparseable/zero tenant) can only ever fire on a corrupted or hand-crafted row — the sole producer always stores a resolved non-zero ID, and the comment itself concedes a repo-level install relying on it would fail anyway. It mirrors the pull_request processor, so consistency argues for keeping it, but at some point both paths could simplify to "bad tenant ⇒ terminal error" rather than carrying a fallback that mostly can't succeed.
- Two driver branches are untested: the terminal error for a row missing repo/head SHA (defensive — the producer always sets both) and the post-run `runCtx.Err()` retry. The latter is genuinely hard to arrange in a test; not blocking either way.

### Verified (tried to break, couldn't)

- **Enqueue/drive guard symmetry.** The driver re-validates every request-path guard: PR presence (ignore), repo registration (complete + unregistered metric), SchemaBot check-name ownership (complete, and `isSchemaBotAggregateCheckName` covers the per-environment aggregate names in both paths), and installation resolution. Ordering is right: the empty-repo/empty-head malformed case is classified as a terminal *error* before the unregistered-repo check can mislabel it — the exact asymmetry I flagged on the push processor in #780, done correctly here.
- **Retry-vs-terminal classification.** Malformed payload → terminal (poison row can't wedge the queue, tested); unsupported action/event → complete; bootstrap failure → retryable; head-verification GitHub failure → retryable ("uncertainty, not staleness" — tested); confirmed newer head → complete without planning a stale snapshot (tested, including that discovery never runs). The `stale`/`error` metric labels match what the synchronous path's `verifyHeadSHAStillCurrentForPR` records, so the two paths aggregate together.
- **Claim-path reachability.** `FindNext`'s claimable predicate has no event-type filter, so `check_run` rows are actually claimed; routing lands in `processDurableCheckRun` and unsupported events/actions complete rather than sit forever.
- **Post-run lease-loss handling is byte-parity with the pull_request driver**: `runCtx` (lease/shutdown) is checked after a successful dispatch while the bootstrap-bounded `ctx` is not, so a slow-but-successful run isn't misclassified and a partial dispatch stays retryable.
- **Flag-off behavior is untouched** — the durable block early-returns before the synchronous bootstrap, and the sync path (including its own head verification) is byte-identical to before.
- **Dedup and redelivery.** Same-GUID redelivery collapses to one row (tested); `Create`'s reopen-on-terminal branch gives operator Redeliver the recovery semantics the enqueue-failure comment promises.
- **No enqueue-before-guard leaks**: a non-SchemaBot check or guard-rejected delivery never creates an inbox row (tested), so the queue only ever carries work the driver would act on.
- The `rerequested` metrics-allowlist addition is correct — it was previously folded to `unknown`.

### Tests

Eleven scenarios across both layers (HTTP enqueue/ack/dedup/skip + driver claim/complete/retry) — this is the right shape and mirrors the #780 suite. The stale-head test asserting discovery never runs (`fileListCalls == 0`) is a nice touch.

aparajon (approved):

Kiran01bm:
> 🤖 Adversarial correctness review, requested by Armand and performed by his agent. Reviewed at head [9bde161](https://github.com/block/schemabot/commit/9bde1611f91149ea66a4d4c26cac6fecf0d8a333), with the same lens as the #780 review (enqueue/drive guard symmetry, retry-vs-terminal classification, claim-path reachability). Verdict up front: this is a solid slice — the guard symmetry is complete, the retry classification is right everywhere I attacked it, and it even fixes the malformed-row-ordering asymmetry class I flagged on #780's push processor. One real finding, plus small notes.

| # | Concern | Status |
|---|---|---|
| 1 | Shutdown-detach applied only to the `check_run` enqueue producer — `enqueueDurablePullRequest`'s INSERT still aborts on request-ctx cancellation, and the comment's "GitHub never auto-retries a check_run rerequest" implies a pull_request/check_run distinction that doesn't exist | `fixed` — lifted into a shared `enqueueDurableWebhookEvent` helper (detached INSERT, 5s bound) covering both producers; comment rewritten to the real rationale. Build + lint (0 issues) + full `pkg/webhook -race` green |
| n1 | Installation-ID fallback (`effectiveInstallationID`) can only fire on corrupted/hand-crafted rows; both durable processors could simplify to "bad tenant ⇒ terminal error" | `deferred` — tracked as [PRF-5](https://github.com/squareup/personal-kmuddukrishna/blob/main/design-stuff/schemabot-pr-review-findings/task-tracker.md) in the PR-review findings tracker, scoped as a both-paths-at-once slice to preserve processor parity |
| n2 | Two driver branches untested: terminal error for a row missing repo/head SHA, and the post-run `runCtx.Err()` retry | `reply` — reviewer marked non-blocking; the first branch is defensive against a row the sole producer can't create, and the second requires deterministically racing lease loss against a successful dispatch, which this layer can't arrange |

PR #822, opened 2026-07-24
refactor(webhook): give apply gates a retryable-vs-terminal disposition

Kiran01bm:
## Summary

The apply command gates fold transient GitHub/storage read failures into their boolean "blocked" result. A durable driver consuming the apply disposition would therefore treat a transient blip inside a gate as a terminal block and permanently drop the apply — even though the direct PR fetches beside the gates correctly treat the same blip as retryable. This gives the five apply/confirm-scoped gates a three-state contract that separates a terminal merit-block from a retryable evaluation failure.

## What

Five gates return `(blocked bool, err error)` instead of a bare `bool`:
`enforceOpenPR`, `enforcePassingChecks`, `enforceReviewGate`,
`checkPriorEnvironments` (plus its local/GitHub helpers), and
`assertBaseSchemaStillCurrent`.

- evaluation failure (a GitHub/storage read the gate needs) → `(true, err)` — fails closed, posts guidance, and marks the block retryable
- blocked on merits → `(true, nil)` — terminal
- passed → `(false, nil)`

`applyCommandCore` maps a gate's `err` to `retry=true` and a merit block to the terminal answer. The synchronous handlers (`handleApplyConfirmCommand`, `executeApply`) are behavior-preserving: because an eval failure keeps `blocked=true`, they still stop on any block via `blocked, _ :=`.

`enforcePRCommandActorAuthorization` is intentionally **not** in this PR — it is the one gate also called from plan/rollback/control/unlock, so it is carved into a follow-up to keep this change scoped to the apply/confirm path.

## Why

This is the prerequisite for converting the durable webhook driver to consume the apply disposition: without it, a durable `issue_comment` apply would fail closed on a transient GitHub outage and never re-drive.

## Before / after
```
before: gate read fails ─▶ return true (blocked)
applyCommandCore ─▶ terminal ─▶ durable driver drops the apply
```

```
after:  gate read fails ─▶ return (true, err)
applyCommandCore ─▶ retry=true ─▶ durable driver re-drives
gate blocks on merits ─▶ (true, nil) ─▶ terminal (unchanged)
```

### DIFF
--- a/pkg/webhook/apply_execute.go
+++ b/pkg/webhook/apply_execute.go
@@ -74,7 +74,10 @@ func (h *Handler) executeApply(
 		h.releaseApplyLockIfIntentUnchanged(ctx, repo, pr, database, dbType, environment, expectedPendingPlanID, "final stale-schema rejection")
 		return
 	}
-	if rejected := h.assertBaseSchemaStillCurrent(ctx, client, repo, pr, installationID, schemaResult, freshPRInfo, environment, requestedBy, actionName); rejected {
+	// executeApply is void; a base-schema evaluation failure only matters to a
+	// durable driver, so the error is discarded here (the gate has already
+	// logged it and posted user guidance) and any rejection releases the lock.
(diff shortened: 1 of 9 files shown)


Kiran01bm:
packaged all of it in pull/823

PR #824, opened 2026-07-24
refactor(webhook): give the actor-authorization gate a retryable-vs-terminal disposition

Kiran01bm:
## Summary
Folds the shared PR-command actor-authorization gate into the same retryable-vs-terminal contract the apply/confirm gates already use, so a durable driver retries on a transient authorization-lookup failure instead of permanently dropping the command.

## What
`enforcePRCommandActorAuthorization` now returns `(blocked bool, err error)`:

| Outcome | Return | Meaning |
|---|---|---|
| team-membership / config read failure | `(true, err)` | retryable — fails closed, posts an authorization-unavailable comment |
| not authorized / unconfigured database | `(true, nil)` | terminal merit block |
| authorized / disabled | `(false, nil)` | passes |

The two already-error-bearing cores (`applyCommandCore`, `applyConfirmCommandCore`) wire the returned error into their retry disposition. The five synchronous callers (plan, control, rollback, rollback-confirm, unlock) explicitly discard the disposition and keep their existing fail-closed terminal-block behavior; the gate logs evaluation errors itself.

## Why
Previously a GitHub team-membership blip was folded into the boolean "blocked" result, so a durable apply driver would classify it terminal and permanently drop the apply — even though the `FetchPullRequest` call beside it correctly treats the same blip as retryable.
```
before:  authz lookup fails ──► return true (blocked)
│
applyCommandCore ───────────┴──► terminal ──► apply dropped

after:   authz lookup fails ──► return (true, err)
│
applyCommandCore ───────────┴──► retryable ──► driver retries
```
This is WH-9c-2a. WH-9c-2b will plumb the disposition through the five synchronous caller signatures once those paths go durable.

### DIFF
--- a/pkg/webhook/actor_authorization.go
+++ b/pkg/webhook/actor_authorization.go
@@ -45,6 +45,17 @@ func (h *Handler) actorAuthorizationClient(
 	return client, false
 }
 
+// enforcePRCommandActorAuthorization gates a PR command on the actor's
+// authorization and reports a retryable-vs-terminal disposition:
+//
+//   - (false, nil) — the actor is authorized, or authorization is disabled.
+//   - (true, nil) — a terminal merit block: the actor is not authorized, or the
+//     database is not configured on this instance. Retrying will not change the
+//     outcome, so a durable driver should stop.
+//   - (true, err) — the authorization decision could not be evaluated (a GitHub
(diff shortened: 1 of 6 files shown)


Kiran01bm:
packaged all of it in pull/823

PR #839, opened 2026-07-24
feat(github): disclose direct-execution changes and require manual confirmation

aparajon:
**Why this matters:** Direct-execution statements are blocking, not revertible, and have no cutover — semantics an operator must consent to explicitly, not discover mid-apply. The PR workflow is where that consent happens.

**What it does:** Plans whose statements the direct execution policy routes to native MySQL DDL render a ⚙️ disclosure on the plan comment and on the locked apply comment, and the automatic apply always downgrades to the two-step manual confirmation regardless of `-y` — confirming against the disclosing comment is the operator's consent. `--defer-cutover` is rejected on an all-direct plan instead of being silently ignored (there is no cutover to defer); on a mixed plan it applies to the engine-driven statements only. Integration tests cover the full apply → downgrade → apply-confirm flow landing a primary-key reshape on the target, and the defer-cutover rejection.

Completes the three-PR chain (blocked-apply gate → direct routing → this): SchemaBot decides the execution path itself, refused-but-necessary changes become possible under an explicit policy with explicit operator consent, and everything else stays exactly as safe as before.


Kiran01bm:
🤖 Review findings - created by Kiran's code review agent - for pull/839, deeb4a9.

**Verdict:** 4 findings — the consent pause itself is fail-closed and works end-to-end, but the doc's central re-plan claim doesn't match the code, and the rollback flow executes direct-routed DDL with no disclosure or gating at all.

### Findings

1. **The rollback flow executes direct-routed DDL with zero disclosure and no gating.** A rollback plan goes through [`ExecuteRollbackPlanForApply`](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/rollback.go#L182) — the standard engine path, so its statements get real `direct`/`blocked` verdicts — but the rollback comment ([rollback.go:254](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/rollback.go#L254)) never populates `DirectChanges`, so the ⚙️ disclosure this PR builds never renders there. Then [`handleRollbackConfirmCommand`](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/rollback.go#L303) calls `h.service.ExecuteApply` directly ([rollback.go:420](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/rollback.go#L420)), bypassing both the `applyCommandCore` gate and the `executeApply` gates entirely. Failure scenario: an operator consents to a small-table PK reshape through the new flow, it lands, and they roll it back — the reverse of a PK reshape is itself a PK reshape, so the rollback plan carries a refused→direct statement; the rollback comment shows plain SQL with no ⚙️ section, and `rollback-confirm` runs blocking, non-revertible native DDL under a consent flow that never disclosed direct execution. Rolling back a direct apply is the *most likely* producer of a direct rollback plan, so this isn't a corner case of the feature — it's its own undo path. (Blocked verdicts on rollback plans are similarly not rejected up front; they dispatch and fail late.)

2. **The consent doc's re-plan claim is wrong: the manual confirm path has no direct gate, and the new `executeApply` gate is dead code on the automatic path.** [docs/direct-execution.md:141-143](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/docs/direct-execution.md#L141-L143) claims the stop-and-confirm "applies again on the pre-execution re-plan, so a plan that gains direct changes between the confirm being requested and the apply executing still stops". That window only exists on the manual confirm path — and there `executeApply` is called with `storedPlan = nil` ([apply_handlers.go:625](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L625)), so the new gate `storedPlan != nil && len(planResp.DirectChanges()) > 0` ([apply_execute.go:141](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L141)) is skipped by design (it must be, or confirm could never execute), and so is the DDL-drift comparison at [apply_execute.go:127](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L127). On the automatic path (the only non-nil caller, [apply_handlers.go:400](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L400)), the gate essentially can't fire either: `applyCommandCore`'s own gate at [apply_handlers.go:356](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L356) already returned for any direct plan, refusal is a pure function of DDL text (an identical-DDL re-plan can't gain a direct verdict), and changed DDL hits the drift gate at :127 first. The genuinely uncovered window: plans diff the desired schema against the **live target**, so an out-of-band change to the target between `apply` and `apply-confirm` — at unchanged PR HEAD, which `assertPlanStillCurrent` can't see — makes the confirm-time re-plan produce a refused→direct statement that no disclosure ever showed, and confirm executes it against consent given to a different plan. Worst variant: the plan-load-failure downgrade at [apply_handlers.go:378](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L378) coaches `apply-confirm` with no ⚙️ section at all. Fix: on confirm, compare the re-plan's direct set against the disclosed set of the pinned plan (`PendingPlanID`), or reword the doc bullet to describe the automatic path only.

3. **`executeApply`'s `--defer-cutover` gate is unreachable on the automatic path, and on the confirm path it destroys the pending confirmation while coaching a retry that can no longer succeed.** At [apply_execute.go:155](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L155), with `storedPlan != nil` an all-direct plan always has `DirectChanges() > 0`, so the direct gate at :141 shadows it (and the front-door gate at [apply_handlers.go:285](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L285) rejected it pre-lock anyway). The only reachable case is `apply-confirm --defer-cutover` on an all-direct plan — where it fires, releases the apply lock ([apply_execute.go:160](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L160)), i.e. discards the pending confirmation, and then [`msgDeferCutoverAllDirect`](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/plan.go#L659) tells the operator "Re-run without the flag" — which now answers "no pending confirmation" and forces a restart from `apply`. Not unsafe (fails closed), but the copy is wrong for the only path that can reach it, and the path has no test.

4. **`executeApply`'s direct-downgrade comment drops the original command's flags and lock state.** The :141 branch rebuilds `commentData` from `buildPlanCommentData` without carrying `AllowUnsafe`/`DeferCutover`/`SkipRevert`/`LockOwner` ([apply_execute.go:144-148](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L144-L148)), unlike the `applyCommandCore` branch, so on an unsafe plan the coached `apply-confirm` command omits `--allow-unsafe` and bounces off the unsafe gate. Mitigating: per finding 2 this branch is effectively unreachable today, and it mirrors the same pre-existing drift-branch behavior — but if the gate is kept as defense-in-depth, its comment should be right.

Could a direct plan slip through to execution without the pause — or, conversely, could the downgrade strand the PR half-locked? Verified fail-closed on both sides. The core gate at [apply_handlers.go:356-373](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L356) fires before any execution, keeps holding the apply lock, pins `PendingPlanID` to the disclosed plan, and stores an `action_required` check record that blocks merge — and `storeApplyPlanCheckRecord` errors are log-and-continue, which leaves the gate still blocking (fail-closed), matching both sibling downgrade branches verbatim. The merge/close webhook only cleans up locks; nothing auto-applies a pending confirmation. The advertised recovery works end-to-end: `TestE2EDirectPlanDowngradesToConfirmThenApplies` drives apply → downgrade comment with ⚙️ disclosure → `apply-confirm` → landed PK reshape, and the confirm path re-verifies HEAD (`assertPlanStillCurrent`, [apply_handlers.go:620](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L620)) and rejects any re-plan that resolved to blocked ([apply_execute.go:115](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L115)).

### Notes (not blocking)

- **No consent surface outside the PR workflow.** The CLI and HTTP transports render no direct-execution disclosure, and the CLI silently ignores `--defer-cutover` on an all-direct plan where the webhook now rejects it. Likely deliberate scoping — worth one line in docs/direct-execution.md saying consent is PR-workflow-only.
- **`shardedDirectChanges` grouping fragments on real Strata keyspaces.** It groups by `(table, reason)` ([plan.go:665](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/plan.go#L665)), but the reason embeds a per-shard row estimate ([engine/spirit/direct.go:205-206](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/engine/spirit/direct.go#L205-L206)), so shards virtually never group and the disclosure degenerates to one bullet per shard; the test hides this by giving both shards identical "~40 rows" reasons. Dead today — the config gate restricts the policy to plain `mysql` ([api/config.go:2519](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/api/config.go#L2519)) — but latent for Strata, and the blocked view has the same issue.
- **Third line-for-line clone of the shard-grouping helper** ([plan.go:623](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/plan.go#L623)/[:665](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/plan.go#L665)/[:701](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/plan.go#L701)) — AGENTS.md's extract-at-3+ rule now applies. Relatedly, [`DirectChanges()`](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/apitypes/apitypes.go#L527) is only ever consumed as `len(...) > 0` and double-counts sharded plans (namespace + per-shard views); a `HasDirectChanges() bool` mirroring [`HasBlockedChanges()`](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/apitypes/apitypes.go#L503) would match convention and remove the overcount footgun.
- **The downgrade reason is a literal at three sites** ([apply_handlers.go:359](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L359), [apply_execute.go:146](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_execute.go#L146), [templates/preview.go:111](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/templates/preview.go#L111)) while this PR's other repeated operator-facing string got a shared const (`msgDeferCutoverAllDirect`).
- **Stale doc comments:** the `ExecutionMode` field doc still describes only `"blocked"` ([apitypes.go:668-670](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/apitypes/apitypes.go#L668-L670)) though this PR edits the constant block right below it; same for `ModeReason` in [storage/types.go:336](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/storage/types.go#L336); and `DirectChangeData`'s doc says "routes to native MySQL DDL" ([templates/plan.go:43](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/templates/plan.go#L43)) though `directConsentCopy` is deliberately engine-keyed with a non-MySQL fallback.
- **Test gaps:** both `executeApply` gates are untested (and per findings 2–3 mostly unreachable — the integration test's apply is stopped by the `applyCommandCore` gate before `executeApply` is reached); mixed-plan `--defer-cutover` pass-through ("applies to the engine-driven statements only") has no end-to-end test; `apply-confirm --defer-cutover` on an all-direct plan — the one reachable path of the :155 gate, including its lock release — is untested.
- **`buildPlanCommentData` drops namespace-level direct entries whenever shard-level entries exist** (if/else at [plan.go:840-857](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/plan.go#L840-L857)). Harmless today (dead with the mysql-only config gate) and it mirrors the blocked view, but the preference is silent.
- Regenerating the docs TOC flags drift in 5 *other* docs — pre-existing on the base branch, not this PR's files; worth a separate hygiene PR.

### Verified correct

- CI fully green on deeb4a9a; `go build` and the unit suites (`pkg/webhook/templates`, `pkg/apitypes`, webhook direct/defer-cutover tests) pass locally at this head.
- TEMPLATES.md is byte-identical to generator output at this head (rebuilt `bin/schemabot`, ran `scripts/update-templates.sh`, empty diff), and docs/direct-execution.md's TOC is in sync.
- Consent bullets 1, 2, and 4 of the new doc section match the code: exact SQL is shown in the plan block, the apply stops at a locked comment holding the lock, and `--defer-cutover` is rejected on all-direct plans at both front doors (pre-lock at [apply_handlers.go:285](https://github.com/block/schemabot/blob/deeb4a9a91b229214ab38360327aec63d04a2ab2/pkg/webhook/apply_handlers.go#L285)) with the mixed-plan footer rendered.
- The `applyCommandCore` direct gate catches verdict-only drift that `ddlMatchesStoredPlan` can't see (same DDL text, changed policy/size verdict) — real defense the drift gate lacks.
- `AllChangesDirect` edge cases (empty plan, mixed, sharded, nil entries) are unit-tested; `DirectExecution()` is nil-safe and mirrors the `EngineBlocked()` precedent, with the mirror constant `executionModeDirect` matching the apitypes value.
- `DatabaseType` is always populated from the schema config, so `directConsentCopy` keying can't silently fall through for registered engines; the Strata/postgres copy branches are deliberate forward-looking dead code (config gate restricts the policy to `mysql`), and the conservative engine-neutral fallback errs toward over-warning.
- `ModeReason` is engine-generated (no user input), so its interpolation into the comment markdown isn't an injection surface.
- Multi-env comments populate per-env `DirectChanges` through `buildPlanCommentData` in production code, not just in the template test's hand-built data.
- `AutoConfirmDowngradeReason` is a pre-existing field reused consistently with both existing downgrade branches (drift, plan-load-failure).

Kiran01bm:
also wrt postgresql - we're good

**No direction change — #839 closes the loop, and it resolves RF-5 in intent at the platform layer.** The research docs' RF-5 invariant (17-invariants.md:249-252) demands that lock-heavy native DDL run only under typed, per-invocation operator consent. #839 delivers the platform half: consent is structurally per-invocation (the pause pins `PendingPlanID`, the confirm is scoped to that pinned plan, and a re-plan that drifts in DDL text or resolves to blocked re-pauses), the disclosure is pinned to the consent comment (exact SQL, per-table impact copy, "not revertible" warning), and there's a double gate (plan-time verdict from #886 + apply-time pause from #836/#839). An adopting PG engine inherits all of this for free — pg-sprite only supplies the verdicts.

Two respects in which #839's consent is *weaker* than RF-5 as written, both acceptable and worth recording rather than fighting: the confirm token is the generic `apply-confirm` command, not a statement-typed token (consent binds to the plan via `PendingPlanID`, not to the DDL text itself); and consent is plan-scoped, not statement-scoped (one confirm covers every direct statement in the plan, and on Strata would cover every shard). The review's finding 1 (rollback path executes direct DDL with no disclosure) and finding 2 (confirm-window target drift is uncovered) are the concrete gaps between "per-invocation in intent" and "per-invocation guaranteed" — track whether upstream closes them, because RF-5's resolution note should cite whatever the final shape is.

aparajon:
🤖 Status: findings 2–4 are fixed on this PR as new commits; finding 1 is queued as its own follow-up.

- **Finding 3** (d7be53d9): the confirm-time `--defer-cutover` rejection no longer releases the lock — the pending confirmation survives, and the copy now coaches re-running `apply-confirm` without the flag. The previously-untested path is now driven end-to-end in the downgrade integration test: rejection preserves the owner and pinned `PendingPlanID`, leaves the target untouched, and a bare re-run executes.
- **Finding 4** (d7be53d9): both auto-confirm downgrade branches (drift and direct — the drift branch had the same gap) now share a helper that carries `--allow-unsafe`/`--defer-cutover`/`--skip-revert` and the lock owner, so the coached command can't bounce off the unsafe gate. The gate itself stays as defense-in-depth; per your finding 2 analysis it's effectively unreachable, so it's covered by the helper's use on the reachable drift path rather than a dedicated test.
- **Finding 2** (1db32319): doc reworded to what the code does — confirm re-verifies the PR head against the confirmed plan and rejects blocked re-plans; the direct-changes downgrade guards the automatic path. The genuinely uncovered window you identified (out-of-band target change between apply and apply-confirm producing an undisclosed direct statement) is real and folded into the follow-up below, where the confirm-time re-plan's direct set gets compared against the disclosed set of the pinned plan.
- **Finding 1**: agreed this is the strongest one — the rollback path is the most likely producer of a direct plan and it bypasses the consent flow entirely. It needs rollback-comment disclosure, up-front rejection of blocked rollback plans, and gating on `rollback-confirm` — too much to bolt onto an approved PR, so it's the next PR after this merges, alongside the plan-load-failure downgrade note.

*— Claude Code (Claude Fable 5)*

